PR #51, opened 2019-10-15
Add a docker compose setup for development

davidfischer:
- Run Postgres, Celery, Redis, and NPM/Webpack with one command
- A few settings required refactoring to run this

Side note: it's possible something similar could work for Read the Docs itself although running Docker in Docker apparently requires some trickery.

## Testing

```bash
$ make dockerbuild   # this can take a while
$ make dockerserve
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,40 +6,22 @@ version: 2
 jobs:
   build:
     docker:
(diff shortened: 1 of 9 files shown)


Sunshine-818:
[]()

PR #70, opened 2019-12-04
Add Sentry for error monitoring

davidfischer:
## Summary
- Lower CPU target tracking from 60% → 50% so ASG scales out before instances get saturated
- Raise max web instances from 6 → 10 for more headroom during traffic spikes
- Raise min web instances from 2 → 3 to maintain baseline capacity
- Reduce gunicorn workers from 6 → 4 per instance to reduce CPU contention on 2 vCPU c6i.large boxes
- Remove unused `WEB_CONCURRENCY` env var from all supervisor configs

## Context
We observed instances hitting load averages of 6+ on 2 vCPU boxes, with all 6 gunicorn workers pinned at ~33% CPU each. The scaling policy wasn't adding instances fast enough (60% target) and the max of 6 didn't leave enough room.

With 4 workers on 2 cores, each worker gets ~50% of a core instead of ~33%, so individual requests complete faster even under load. The higher min (3) keeps baseline capacity at 12 total workers (4×3), same as before (6×2). The 50% CPU target keeps roughly half the workers idle on average, leaving headroom for traffic spikes while new instances boot.

## Deploy notes
- Terraform changes (scaling policy, min/max size) can be applied immediately
- Gunicorn worker change requires an AMI rebuild + rolling deploy

### DIFF
--- a/config/settings/production.py
+++ b/config/settings/production.py
@@ -94,6 +94,18 @@ CELERY_BROKER_URL = env("CELERY_BROKER_URL")
 CELERY_RESULT_BACKEND = CELERY_BROKER_URL
 
 
+# Sentry settings for error monitoring
+# https://docs.sentry.io/platforms/python/django/
+# --------------------------------------------------------------------------
+SENTRY_DSN = env.bool("SENTRY_DSN", default=None)
+if SENTRY_DSN:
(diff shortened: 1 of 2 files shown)


humitos (approved):
LGTM!

PR #150, opened 2020-06-30
Allow publishers to have paid campaigns only

davidfischer:
This sets an admin-only flag to set a publisher to only allow paid campaigns.


davidfischer:
> the idea here being that we shouldn't show house or community ads on these?

Correct. I'm sure some publishers aren't looking for that.

PR #151, opened 2020-07-01
Version v0.1.10

davidfischer:


davidfischer:
> At some point we should move to 1.0 :D

We will. Nobody looks at the versions though =)

PR #156, opened 2020-07-02
Add the basic embed code to the adserver dashboard for publishers.

ericholscher:
This is just a start,
but should make things a bit easier for onboarding.


davidfischer:
The failure here is only because it reduced our coverage. I'll see if I can get a fix in there.

adserver/templates/adserver/base.html
```diff
@@ -73,12 +73,21 @@ <h6 class="text-muted">{{ advertiser }}</h6>
...
             <ul class="nav flex-column mb-2">
+
               <li class="nav-item">
                 <a class="nav-link" href="{% url 'publisher_report' publisher.slug %}">
                   <span class="fa fa-bar-chart fa-fw mr-2 text-muted" aria-hidden="true"></span>
                   <span>{% trans 'Reports' %}</span>
                 </a>
               </li>
+
+              <li class="nav-item">
```
davidfischer, line 84: This block should only appear if the publisher allows `unauthed_ad_decision`. Otherwise, it won't work because the JSONP request will be rejected.

adserver/templates/adserver/reports/publisher_embed.html
```diff
@@ -0,0 +1,49 @@
...
+  <section>
+    <div>
+      <h2></h2>
+      <div class="row col-md-8">
+
+        <p>
+          This code will allow you to get started with showing ads on your site.
+        </p>
+
+        <pre>
```
davidfischer, line 30: You might get something a little better if you do `<pre><code>...</code></pre>`: https://getbootstrap.com/docs/4.1/content/code/

PR #157, opened 2020-07-03
Add django-extensions so we can use `manage.py shell_plus`

ericholscher:


requirements/development.txt
```diff
@@ -3,6 +3,7 @@
 
 # Useful debugging tools
 django-debug-toolbar<2.0
+django-extensions<2.2
```
davidfischer, line 6: This is already in https://github.com/readthedocs/ethical-ad-server/blob/master/requirements/base.txt.

PR #159, opened 2020-07-03
Minimal changes to support default keywords for a publisher

ericholscher:


adserver/models.py
```diff
@@ -123,6 +131,12 @@ def __str__(self):
     def get_absolute_url(self):
         return reverse("publisher_report", kwargs={"publisher_slug": self.slug})
 
+    @property
+    def keywords(self):
```
davidfischer, line 135: This function (or the model validator) needs to be fleshed out a bit more. For example, this is what .org does: https://github.com/readthedocs/readthedocs.org/blob/505be8cc8fabc0dbdf290336dd263c887ede459b/readthedocs/projects/tag_utils.py#L12-L29

- Lowercase the tags
- Slugify the tags
- Replace underscores
- `from taggit.utils import _parse_tags` also does some smart things like detecting that the user used spaces instead of commas. That's less required but nice.

adserver/models.py
```diff
@@ -552,6 +566,14 @@ def show_to_keywords(self, keywords):
         If *any* keywords are in the excluded list, it should not be shown.
         """
         keyword_set = set(keywords)
+        # Add default keywords from publisher
```
davidfischer, line 569: The publisher keywords should probably just be added to the keywords list in the decision engine here: https://github.com/readthedocs/ethical-ad-server/blob/eb2b99baf24786ebca8eab7b2cc18e6918ce64c0/adserver/decisionengine/backends.py#L38

adserver/models.py
ericholscher: I've done some basic cleanup here, but we're not exposing this to users quite yet, so I think its fine to be a bit rough.

PR #161, opened 2020-07-04
Record views for ad network for now

ericholscher:
We will tune this over time, but lets get more data for now.

### DIFF
--- a/adserver/models.py
+++ b/adserver/models.py
@@ -1029,7 +1029,8 @@ class Advertisement(TimeStampedModel, IndestructibleModel):
         """
         self.incr(VIEWS, publisher)
 
-        if settings.ADSERVER_RECORD_VIEWS:
+        # TODO: Find a better way to record explicit ad server views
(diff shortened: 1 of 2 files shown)


adserver/models.py
```diff
@@ -1029,7 +1029,8 @@ def track_view(self, request, publisher, url):
         """
         self.incr(VIEWS, publisher)
 
-        if settings.ADSERVER_RECORD_VIEWS:
+        # TODO: Find a better way to record explicit ad server views
+        if settings.ADSERVER_RECORD_VIEWS or "readthedocs" not in publisher.slug:
```
davidfischer, line 1033: I do want to note that I view this as temporary and we will need a better solution. This sort of breaks the setting for anyone else who uses our ad server.

ericholscher: Totally -- I think an explicit flag on the publisher is probably the correct option.

davidfischer: I think that's a better option, overall. Maybe we have the global setting and it we have a publisher specific setting that is default=False.

PR #162, opened 2020-07-04
Use any python3 for precommit

ericholscher:
This was breaking on my local machine with python3.7


ericholscher:
It would be particularly neat if this used the docker-compose env, but I don't think that makes sense really.

ericholscher:
Yea, I'm not actively runniing pre-commit on RTD because it breaks randomly, but trying to keep it on here.

PR #164, opened 2020-07-06
Have an explicit flag to record publisher views

davidfischer:
This is a different solution for https://github.com/readthedocs/ethical-ad-server/pull/161 that adds an explicit flag so we can turn it on and off on a per-publisher basis.


davidfischer:
It defaults to off so we'll have to explicitly turn it on everywhere we want it.

PR #168, opened 2020-07-08
Small admin cleanup

ericholscher:


adserver/admin.py
```diff
@@ -210,6 +213,7 @@ class AdvertisementsInline(AdvertisementMixin, admin.TabularInline):
     model = Advertisement
 
     can_delete = False
+    list_per_page = 50  # make page load a bit faster
```
davidfischer, line 216: I don't think there will be a ton of 50+ ad flights. I also don't think this actually has any effect. It [isn't one of the fields shared with ModelAdmins](https://docs.djangoproject.com/en/2.2/ref/contrib/admin/#inlinemodeladmin-options).

adserver/admin.py
```diff
@@ -146,6 +148,7 @@ class AdvertisementAdmin(RemoveDeleteMixin, AdvertisementMixin, admin.ModelAdmin
 
     model = Advertisement
     save_as = True
+    list_per_page = 50  # make page load a bit faster
```
davidfischer, line 151: This is fine for now. This admin would be *very* fast if we removed the aggregation across impressions. Maybe we should just do that.

adserver/admin.py
davidfischer, line 216:
```suggestion
```

ericholscher: Yea, I meant to put it on the AdvertisementAdmin itself I think.

adserver/admin.py
```diff
@@ -476,7 +479,7 @@ class AdBaseAdmin(RemoveDeleteMixin, admin.ModelAdmin):
     )
     list_display = readonly_fields[:-3]
     list_select_related = ("advertisement", "publisher")
-    list_filter = ("is_mobile",)
+    list_filter = ("is_mobile", "publisher")
```
davidfischer, line 482: I just realized this will not cascade down to clicks because `list_filter` is overridden there. We should either add it there or remove the override. I say we just remove the override.

davidfischer: I'll add it there and we can remove the override.

adserver/admin.py
```diff
@@ -502,7 +505,7 @@ class ClickAdmin(AdBaseAdmin):
 
     # Browser Family and OS Family are not in the ``ViewAdmin.list_filter``
     # because they require a ``SELECT DISTINCT`` across the whole table
-    list_filter = ("is_mobile", "is_bot", "browser_family", "os_family")
+    list_filter = ("is_mobile", "publisher", "browser_family", "os_family")
```
davidfischer, line 508: I replaced the `is_bot` filter which is totally useless anyway. We already exclude bots from counting toward views and clicks anyway. That field can probably just be deleted on the model.

davidfischer:
This is good to merge and I'll merge it unless there are objections in a bit.

PR #169, opened 2020-07-08
Add additional filtering to reporting

ericholscher:
This adds:

* The campaign_type filter to Publishers
* The revenue_share_percentage to All Publishers

### DIFF
--- a/adserver/models.py
+++ b/adserver/models.py
@@ -163,7 +163,13 @@ class Publisher(TimeStampedModel, IndestructibleModel):
             return return_keywords
         return []
 
-    def daily_reports(self, start_date=None, end_date=None, campaign_type=None):
+    def daily_reports(
+        self,
+        start_date=None,
+        end_date=None,
+        campaign_type=None,
+        revenue_share_percentage=None,
+    ):
         """
         Generates a report of clicks, views, & cost for a given time period for the Publisher.
 
@@ -183,6 +189,15 @@ class Publisher(TimeStampedModel, IndestructibleModel):
             impressions = impressions.filter(
                 advertisement__flight__campaign__campaign_type=campaign_type
             )
+
+        revshare_options = set(
+            str(pub.revenue_share_percentage) for pub in Publisher.objects.all()
+        )
+        if revenue_share_percentage and revenue_share_percentage in revshare_options:
+            impressions = impressions.filter(
+                publisher__revenue_share_percentage=revenue_share_percentage
+            )
+
         impressions = impressions.select_related(
             "advertisement", "advertisement__flight"
         )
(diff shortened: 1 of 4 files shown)


adserver/models.py
```diff
@@ -163,7 +163,13 @@ def keywords(self):
             return return_keywords
         return []
 
-    def daily_reports(self, start_date=None, end_date=None, campaign_type=None):
+    def daily_reports(
+        self,
+        start_date=None,
+        end_date=None,
+        campaign_type=None,
+        revenue_share_percentage=None,
```
davidfischer, line 171: I don't think we want this here. This report is already specific to a single publisher. If we want to limit the publishers we are running this across, that should be done before this method is called.

Since this figure doesn't differ for a publisher, do we want this?

adserver/templates/adserver/reports/publisher.html
```diff
@@ -14,6 +14,18 @@
   <li class="breadcrumb-item active">{{ publisher }}</li>
 {% endblock breadcrumbs %}
 
+{% block additional_filters %}
```
davidfischer, line 17: Is the reasoning behind this to see non-paid campaigns for specific publishers? This filter is exposed to publishers. Do we think that's important to them?

ericholscher: Yea, at least for now I'm curious to see the paid vs. non-paid views. But maybe once we're fully spun up and turn `paid ads only` on, it won't be important?

adserver/models.py
ericholscher: It seems all the filtering is done in this function, so for the `All publishers` report, it seems like we need to pass it in?

adserver/templates/adserver/reports/publisher.html
davidfischer: I think it's fine for now. I suspected that most publishers would want to restrict ads shown on their property to paid only. However, maybe that isn't the case. Maybe rather than `Publisher.paid_ads_only` we need multiple options to allow paid, allow house, etc.

adserver/models.py
davidfischer: This function is per publisher. We need some filtering in the function because there could be both paid and house campaigns shown on the publisher. So if you want to filter by campaign type, that should be done on a per publisher basis because there are both paid and house impressions across the publisher. However, every impression for a publisher has the same revshare percentage so that could be done before it gets to here.

To make this concrete, [this is what I'm proposing](https://github.com/readthedocs/ethical-ad-server/compare/improve-filtering-options...davidfischer/revshare-filtering-alternative).

ericholscher: Ah yea, thats fine -- I was just using the existing pattern.

davidfischer: I turned it into a PR #171.

Issue #163, opened 2020-07-06
Ad types and existing advertisers and advertisements


davidfischer:
## The problem

One problem we have is that we have hundreds of existing ads in our system and we want to make it easy for advertisers to apply an existing ad across our new fledgling ad network.

The way that our current system is structured is that when an ad is requested by a site on the network (or Read the Docs), the exact ad type is specified. So a site will explicitly request an image+text ad. This is great except that then advertisers need to create 2 versions of the same ad: one with an image and one without even though they're basically the same. The other problem is that we have a separate ad type for ad network image+text ads from RTD image+text ads so an advertiser would need 2 versions of exactly the same ad (image+text for RTD, image+text for network) and then a 3rd version that is text-only. Long story short, for an existing advertiser to just show ads across our network, we need to duplicate their ads which is not desirable.

## Background

To give a bit of background, almost all of the ads in our existing system have an ad type that is Read the Docs specific. There are 3 RTD ad types:

* Sidebar ads: 80 characters + 240*180 image
* Footer ads: 300 characters + 240*180 image
* Fixed footer ads: 80 characters, no image

The 300 character footer ad type is deprecated and there are only a few of them but it is used on RTD. I don't think we ever want to use this across our ad network. Most of the publishers on our network want either a simple ad with an image or just a text ad. These two types are analogous to the Sidebar ad and the Fixed Footer ad on Read the Docs.

## Solutions

These are a couple different options. They are similar but we only need to do one of them.

### Move display logic into the ad client

Publishers will continue to express a preference for a text-only or text+image ad. However, the ad server would need a modification to send down the components of the ad like the text and the image and the client could assemble the HTML based on the publisher preference. In this case, the ad type on the ad server would be the same regardless of whether the ad is displayed with an image or without.

With respect to our existing ad types, I would suggest:

* Change the name of the sidebar ad to something like "Text and optional Image". Combine the RTD sidebar ad type and the "Ad network generic image v1" type. This ad type will be displayable either at text+image or text-only based on the publisher preference. On RTD, this will continue to be displayed only in the sidebar.
* Remove the "Ad network generic text" ad type. Whether to display an image+text or text-only will be by the publisher preference.
* The fixed footer type and the footer ad types will be RTD specific and not used elsewhere. They will continue to exist and simply be marked deprecated (the 300 character footer ad already is)

### Ad server accounts for the preference in addition to/instead of ad type

[...]

PR #170, opened 2020-07-08
Convert Advertisement.ad_type to many-to-many

davidfischer:
- Adds a data migration to migrate the FK -> M2M
- Does not yet remove the FK field which could be removed later
- Validation, which relies on ad types,  was moved from a validator into a form (which the admin uses as well). In some ways, handling via a form was better.
- Validation now checks that ad is valid across all selected ad types. Validating that a text-only ad has an image is ignored.

This allows an ad to have multiple ad types. The same ad could then be displayed in different ways
based on the ad types chosen. The decision API's requested ad type will control which ad type is used for display.

Closes https://github.com/readthedocs/ethical-ad-server/issues/163


davidfischer:
> I will get this running locally to do a bit more testing as well prior to deployment, just to give it a bit more QA.

Thanks.

> other unintended outcomes

There's a few interesting bits:

- All the validation is moved to the advertisement create/update forms. Both the existing form advertisers use and the admin form use the same validation code, but by creating a model directly this validation could be bypassed. In some ways I like this better because previously there was slightly different validation on create vs. update and now its the same. However, I also liked having validation on the model which isn't really possible with the m2m.
- I do like having default checked ad types. I think this will help publishers. I also think we will need good names/descriptions for these new ad types. That can be done after this is deployed. I'm leaning toward RTD Sidebar checked, Image+Text checked, Text-Only (ad network) checked.
- The migration does *NOT* remove the old `Advertisement.ad_type` yet. My thought process is that we want to run the migration right before we deploy and this will allow having a zero-downtime deploy. I did verify that absolutely 0 places in the code use the old field though. I think we remove it in the deploy after this.

ericholscher:
Other than my local styling issues, this seemed to work great.

PR #179, opened 2020-07-09
Preview new ad types

davidfischer:
## Testing locally

You'll need to grab the latest ad type templates for the ad network ad types from the server. A small update was required due to https://github.com/readthedocs/ethical-ad-client/pull/21 (basically adding `ea-type-text` or `ea-type-image` classes).

Create an image and preview it on the ad detail page, a report, or the ad form.


assets/src/scss/_ad-styles.scss
```diff
@@ -0,0 +1,220 @@
+/* Taken from https://github.com/readthedocs/ethical-ad-client/blob/master/styles.scss */
```
ericholscher, line 1: Seems like it would be better for us to publish these as part of the client, and link them from there. That would keep us up to date with changes.

davidfischer:
> If we're adding styles into the templates to show, will that not also change how the client uses the loaded attribute?

This preview always sets the `loaded` attribute.

assets/src/scss/_ad-styles.scss
davidfischer: Currently, the ad server does not rely directly on the client. Are you proposing we do that? Or are you just proposing that we build the styles from the client into the CSS for the ad server by making the client a dev dependency?

ericholscher: Not sure -- we can deal with it later, but copying & pasting them here feels like it will get out of date, but we probably have a while before that's an issue, and we can deal with it the.

davidfischer: Cool. I removed it and imported it from node. However, we will need to manage the version we link against. Should be better than the first pass though.

davidfischer:
I did a slight refactor here:

- Add the ad client as a dependency so the styles come from there
- Render the ad preview in a custom template tag

PR #180, opened 2020-07-10
Refactor flight metadata view

davidfischer:
~~This shows the first 5 publishers for a flight followed by a message like "and 2 other publishers". This at least exposes the basic information to advertisers. I think once we have a publisher reporting breakdown for advertisers we can link to it for more details like top publishers for a campaign.~~

This PR is now just a simple refactor to have an include for high level flight metadata. It does not show publishers at all. Those will be shown in a new report that needs to be built.


adserver/templates/adserver/includes/flight-metadata.html
```diff
@@ -0,0 +1,73 @@
...
+  {% if flight.end_date %}
+    <dt>{% trans 'Estimated end date' %}</dt>
+    <dd>{{ flight.end_date }}</dd>
+  {% endif %}
+  {% with flight.campaign.publishers.count as publishers_count %}
+    {% if publishers_count %}
+      <dt>{% trans 'Publishers' %}</dt>
+      <dd>
+        <ul>
+          {% for publisher in flight.campaign.publishers.all|slice:":5" %}
```
ericholscher, line 41: Doesn't seem like this is sorted by pageview or weight at all? This seems like a pretty bad thing to expose, since I don't think it will show what advertisers expect all the time, and could cause some confusion. I definitely think it should at least be weighted by impression count.

davidfischer: I definitely think showing the top by weighted impressions is ideal. The question then becomes over what timeframe? I also would like to show this data even for upcoming flights so that advertisers know where their ad will be shown. I think both are useful data. However, I don't want to mislead. Maybe this requires more thought.

ericholscher: > The question then becomes over what timeframe?

The timeframe of the report date filter? 

Upcoming seems a bit trickier, but that feels like a different thing than "where are my ads currently being shown" that is more like "Where are all the possible places my ad might be shown". Both are useful, but I think the first is the most important currently.

davidfischer:
I'm going to hold off on this PR until after we've built something for [grouping publishers](https://github.com/readthedocs/ethical-ad-server/pull/202#discussion_r463300495).

adserver/templates/adserver/includes/flight-metadata.html
davidfischer: > The timeframe of the report date filter?

This same data is shown on the flight detail view which doesn't have any dates associated with it. I think this data will be available exclusively on a per-publisher detail report.

davidfischer:
I updated this PR to just be a straight refactor and updated the description.

PR #181, opened 2020-07-13
JSONP client parity

davidfischer:
- Support keyword targeting
- Support campaign-type restriction

### DIFF
--- a/adserver/api/views.py
+++ b/adserver/api/views.py
@@ -41,16 +41,29 @@ class AdDecisionView(GeoIpMixin, APIView):
     * geography (based on IP)
     * keywords
 
-    .. http:post:: /api/v1/decision/
+    .. http:get:: /api/v1/decision/
 
-        :<json string publisher: **Required**. The slug of the publisher
-        :<json array placements: **Required**. Various possible ad placements where an ad could go
-        :<json array keywords: Case-insensitive strings that describe the page where the ad will go for targeting
-        :<json array campaign_types: Limit the ad results to certain campaign types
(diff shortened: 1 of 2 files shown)


adserver/api/views.py
```diff
@@ -74,6 +74,8 @@ class AdDecisionView(GeoIpMixin, APIView):
             The number and order must correspond to the ``div_ids``.
         :<json string priorities: An optional ``|`` delimited string of priorities for different ad types.
             The number and order matter, applying to ``div_ids`` and ``ad_types``.
+        :<json array keywords: An optional ``|`` delimited string of keywords.
+        :<json array campaign_types: An optional ``|`` delimited string of campaign types.
```
ericholscher, line 78: We should probably define these concepts somewhere (eg. what's a campaign type? Keyword?), if we haven't. I had a user ask what they are recently when I linked them to this.

adserver/api/views.py
```diff
@@ -74,6 +74,8 @@ class AdDecisionView(GeoIpMixin, APIView):
             The number and order must correspond to the ``div_ids``.
```
ericholscher, line 74: We should probably remove the mention of `Supports the same parameters as via POST with the following changes:` now I think, and just include the full content here?

davidfischer:
I updated the documentation and filled out some details and the GET request section specifically.

davidfischer:
I do think a lot of these examples should go into the https://github.com/readthedocs/ethical-ad-client though since that's where it'll be used.

ericholscher:
Docs changes look great. I agree we want more of this content on the client as well 👍

PR #182, opened 2020-07-13
Add `body` to the API response

ericholscher:
This is a text-only version of the ad for now.

### DIFF
--- a/adserver/api/views.py
+++ b/adserver/api/views.py
@@ -68,7 +68,8 @@ class AdDecisionView(GeoIpMixin, APIView):
         :<json string force_campaign: Limit results to ads from a specific campaign
 
         :>json string id: The advertisement slug of the chosen ad
-        :>json string text: The text of the ad
+        :>json string text: **Deprecated**:The HTML text of only the ad.
+        :>json string body: The text of the ad, stripped of any HTML.
         :>json string html: An HTML rendering of the ad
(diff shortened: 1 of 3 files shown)


adserver/api/views.py
```diff
@@ -68,7 +68,8 @@ class AdDecisionView(GeoIpMixin, APIView):
         :<json string force_campaign: Limit results to ads from a specific campaign
 
         :>json string id: The advertisement slug of the chosen ad
-        :>json string text: The text of the ad
+        :>json string text: **Deprecated**:The HTML text of only the ad.
```
davidfischer, line 71: Do we actually want to deprecate this? I don't think we do. I suspect a lot of people (and probably our ad client) will still want to just do something like `$("#div").html(resp.text)`. I just think that having the text be simply a put together version of the body and the headline is ideal.

ericholscher: The client is using the `html` in the response -- which I assume is what most people would want if they are just plopping HTML in?

ericholscher: I'm happy not to deprecate it for now tho.

davidfischer: You're right! Maybe `text` should just be plain text or maybe we should just get rid of it.

ericholscher: Yea, I think people are using it currently, but we should definitely remove once we fully migrate.

PR #184, opened 2020-07-14
Update docs to explain auth on POST request

ericholscher:


docs/user-guide/api.rst
```diff
@@ -20,7 +21,7 @@ Authentication works in one of two ways:
 .. note::
 
     Creating a token for a user can be done in the :doc:`administration interface </user-guide/administration>`
```
davidfischer, line 23: We should probably update this line as well:

Something like:

    Creating an API token can be done under :guilabel:`Username Dropdown > API Token`.

(I'd make a multi-line suggestion but it isn't working due to a deleted line)

ericholscher: We could even link to it if we wanted: https://server.ethicalads.io/accounts/api-token/

PR #185, opened 2020-07-14
Stripe integration

davidfischer:
This is not a complete integration but it does allow creating invoices from the Django admin for a specific advertiser with metadata attached.

- Adds a field `Advertiser.stripe_customer_id`
- Adds a two admin actions for creating an invoice:
  * One is on the advertiser object to create a generic invoice for an advertiser not tied to any flight. This is useful for just topping up an account for an advertiser.
  * The other can be applied to one or more flights. All the flights have to be on the same advertiser. Each flight will be a separate line item on the invoice. This is useful for when an advertiser is running 3 flights (say divided into different regions) and we want to have one invoice for all 3.
- Created invoices are just drafts that can be edited before being sent.


adserver/admin.py
```diff
@@ -65,9 +73,74 @@ class AdvertiserAdmin(RemoveDeleteMixin, admin.ModelAdmin):
...
+        if not settings.STRIPE_SECRET_KEY:
+            messages.add_message(
+                request,
+                messages.ERROR,
+                _("Stripe is not configured. Please set settings.STRIPE_SECRET_KEY."),
+            )
+            return
+
+        flight_start = timezone.now()
+        flight_end = flight_start + timedelta(days=30)
```
ericholscher, line 92: Should we not put this on the actual Flight object, where we could then pull the billing data?

davidfischer: Actually that's a great idea. It doesn't have to be either/or. It could be both and then we'll see what actually gets used by a real workflow. Ideally, each flight is an invoiceitem and the whole thing is an invoice.

davidfischer: My thought process behind both is that as we move from billing for specific flights to topping up an account, we might want it on the advertiser. I'm happy to get both and remove one in the future.

davidfischer:
Added creating an invoice with multiple invoice items from multiple flights.

PR #187, opened 2020-07-15
Order the Ad admin by created date, not slug

ericholscher:
We should probably change the sorting on this model in general,
but I'm not sure why it's `ordering = ("slug", "-live")` so I didn't mess with it.

### DIFF
--- a/adserver/admin.py
+++ b/adserver/admin.py
@@ -68,7 +68,6 @@ class AdvertiserAdmin(RemoveDeleteMixin, admin.ModelAdmin):
     list_display = ("name", "report")
     prepopulated_fields = {"slug": ("name",)}
     readonly_fields = ("modified", "created")
-    ordering = ("-created",)
 
     def report(self, instance):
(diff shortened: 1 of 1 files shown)


davidfischer:
The edit in this PR refers to the Advertiser model but your comment refers to the Advertisement model. I don't think there's a good reason about the ad ordering especially since slug has a unique key on it (I think it didn't in the old system).

ericholscher:
Ugh, I seem to be incapable of editing the proper admin for Adverisements heh

ericholscher:
@davidfischer the creation date should be deterministic, no?

davidfischer:
> @davidfischer the creation date should be deterministic, no?

No it isn't. Two objects can have the same create date. Basically, Django does some odd things in cases where the order isn't perfectly deterministic: https://docs.djangoproject.com/en/2.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.ordering

Issue #188, opened 2020-07-15
Flight targeting to include/exclude mobile traffic


davidfischer:
A few advertisers have asked to exclude mobile traffic. This seems like an easy enough feature to add. I could also see advertisers asking for only mobile traffic but nobody has actually asked that.

PR #192, opened 2020-07-16
Allow targeting mobile or non-mobile traffic

davidfischer:
This allows the targeting parameters to be:

    {"mobile_traffic": "exclude"}

OR

    {"mobile_traffic": "only"}

to exclude mobile traffic or to target only mobile traffic.

Fixes: #188

### DIFF
--- a/adserver/decisionengine/backends.py
+++ b/adserver/decisionengine/backends.py
@@ -3,6 +3,7 @@ import logging
 import random
 
 from django.db import models
+from user_agents import parse
 
 from ..constants import AFFILIATE_CAMPAIGN
 from ..constants import ALL_CAMPAIGN_TYPES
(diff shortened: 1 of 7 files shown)


davidfischer:
> Doesn't need to be in this PR, but I often find myself needing to look at old Flights in the DB or tests in this repo to find targeting syntax.

Agreed 100%. Like RTD itself, we run into the issue of separating out advertiser docs, publisher docs, developer/installation docs, and administration docs.

davidfischer:
At some point, I think we need a smarter interface for managing targeting. It's on the roadmap but just hasn't been a priority.

ericholscher:
I think we'll need to figure it out when we expose it to users, so that'll be the obvious time to do it.

PR #195, opened 2020-07-18
Let publishers control campaign types of advertisers

### LINKED ISSUE #176: Allow publishers to control whether they get house/community campaigns
Currently we have a flag for publishers to allow only paid ads `Publisher.paid_campaigns_only`. Instead of this, I think we want to move to 3 flags -- `allow_affiliate_campaigns`, `allow_community_campaigns`, `allow_house_campaigns`. We don't really do much in the way of affiliate campaigns right now (none on the ad network) and with Covid stopping conferences there aren't many community campaigns currently although there's been lots in the past. However, controlling whether we show house ads is useful. Established publishers may not want this especially if they fallback to another ad network but some publishers may not care or they might allow community ads but not house ones. I don't know if we need a flag to allow paid advertising. I guess it's easy enough to add.

* Add 4 flags to the publisher model.
  * `allow_paid_campaigns` (default true)
  * `allow_affiliate_campaigns` (default false)
  * `allow_community_campaigns` (default true)
  * `allow_house_campaigns` (default true)
* Add a data migration which sets `allow_affiliate_campaigns`, `allow_community_campaigns`, `allow_house_campaigns` to false if `Publisher.paid_campaigns_only`.
* After deploying, turn on affiliate campaigns for RTD itself.
* Once we have the above, we can expose this setting to publishers so they can control this themselves.

davidfischer:
On their own site, publishers should have control over whether they get house or community ads.

Fixes #176

## Screenshot

Screen Shot 2020-07-18 at 11 55 59 AM

**Note:** controlling paid ads is now an admin-only action.


davidfischer:
One thought I had is to make "allow paid campaigns" an admin only setting. This way we can control whether a publisher gets paid ads and turn it on when there integration is reviewed. This also gives us an easy way to disable a publisher in the event of an issue. We could have a note at the top of the publisher screen saying "paid ads are currently disabled, please contact us when you're ready for us to review your site and enable them".

ericholscher:
Yea -- I think we probably need some kind of "Approved" state for publishers, instead of using this as a proxy. I agree though that we shouldn't let users enable paid ads on their site without this, but I'm thinking it probably makes more sense to have logic thats like `filter(revshare=70, approved=True)` when people want things shown on the ad network by default.

adserver/forms.py
```diff
@@ -214,3 +218,45 @@ class Meta:
...
+                ),
+                css_class="my-3",
+            ),
+            Submit("submit", "Save settings"),
+        )
+
+    class Meta:
+        model = Publisher
+        fields = [
+            "allow_paid_campaigns",
```
ericholscher, line 258: Yea -- I'd say this shouldn't be a boolean, but should probably just show the status, and let us change it.

adserver/templates/adserver/publisher/embed.html
```diff
@@ -0,0 +1,39 @@
+{% extends "adserver/base.html" %}
```
ericholscher, line 1: Wonder why git didn't see this as a moved file -- different commits?

adserver/forms.py
davidfischer: Just so we're clear, you're thinking something like `Publisher.status` and the values are "approved" and "not approved"? Would a boolean `Publisher.approved` be better?

ericholscher: I think `allow_paid_campaigns` is fine for now, and we can treat it like `approved` as something that we control. Seems fine to me.

ericholscher: Sorry -- my initial comment was confusing. I just meant that we shouldn't let the user change it, but it should be something only we can change.

davidfischer: Sounds good. I'll remove it from the form and we should be good to go.

davidfischer:
I updated this PR with a few changes:

- Remove publishers ability to set paid ads. This will be an admin only action and part of our approval process.
- The campaign_types parameter to the decision API can only remove campaign types, not add ones the publisher isn't approved for.
- Show a warning to publishers that they aren't approved for paid ads. They will need to contact us for review and approval.

PR #196, opened 2020-07-20
CSRF Cookie tweaks

davidfischer:
- Lower the timeout from 1y -> 30d. This is only used for form checking. A CSRF error could arise if a user opens a form and submits it a month later. Seems unlikely and they could just reload the page and submit.
- Sets the SameSite attribute to strict on the CSRF cookie. The CSRF cookie is set anytime a user loads a form (eg. login form). By setting it to strict, this cookie won't be sent in a 3rd party context. Say a user loads our login form and then sees an ad on another site, it won't send the cookie with the ad load.

We want the session cookie to be sent in 3rd party contexts so that one should stay "lax". This lets us know when a publisher loads their own ads for example and that our link from https://ethicalads.io -> https://server.ethicalads.io/accounts/login/ will redirect correctly if the user is already logged in. We only set the session cookie if somebody is logged into the ad server so that seems correct.

These changes should be fine but we should monitor 403 CSRF errors just in case.


davidfischer:
This was not driven by a user request. The cookie age matches what we have on RTD. Setting strict is something new but I think it's more important here than on RTD itself. This cookie is required for things like the login form to work but it is never necessary for advertising and should head off potential questions.

PR #201, opened 2020-07-27
Allow simple variables in Advertisement.link

davidfischer:
This lets advertisers have simple variables in the landing page URLs for ads:

    https://example.com?utm_source=${publisher}

It handles bad syntax by just not doing the replacement.


davidfischer:
My first pass used a Django template, but then I thought about it and a Django template is more than we need here. We just want simple variable substitution. We don't want `{% if %}`, `{% for %}`, `{% include %}` (a possible security issue), etc.

ericholscher:
True -- seems fine for now.

PR #202, opened 2020-07-28
More admin improvements

davidfischer:
- Use raw_id_fields for most things that are getting large
- filter by advertiser instead of campaign most places

### DIFF
--- a/adserver/admin.py
+++ b/adserver/admin.py
@@ -564,7 +564,7 @@ class CampaignAdmin(RemoveDeleteMixin, admin.ModelAdmin):
     )
     list_filter = ("campaign_type", "advertiser")
     list_select_related = ("advertiser",)
-    raw_id_fields = ("advertiser", "publishers")
+    raw_id_fields = ("advertiser",)
     readonly_fields = ("campaign_report", "total_value", "modified", "created")
     search_fields = ("name", "slug")
 
--- a/adserver/admin.py
+++ b/adserver/admin.py
@@ -254,30 +254,25 @@ class AdvertisementAdmin(RemoveDeleteMixin, AdvertisementMixin, admin.ModelAdmin
         "ecpm",
     )
     list_display_links = ("name",)
-    list_select_related = ("flight", "flight__campaign")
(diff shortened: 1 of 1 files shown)


adserver/admin.py
```diff
@@ -563,6 +564,7 @@ class CampaignAdmin(RemoveDeleteMixin, admin.ModelAdmin):
     )
     list_filter = ("campaign_type", "advertiser")
     list_select_related = ("advertiser",)
+    raw_id_fields = ("advertiser", "publishers")
```
ericholscher, line 567: This actually will make it a good deal harder to work with, won't it? We'll need to select publishers one-by-one to add them to a campaign? 

I think we probably need to automate this rather than change it to raw_id, like with an admin action that adds all ad network publishers (`filter(revshare=70, approved=True)`), or RTD publishers (`filter(revshare=50)`).

davidfischer: > This actually will make it a good deal harder to work with, won't it? We'll need to select publishers one-by-one to add them to a campaign?

That's true but it'll be much easier to duplicate the setup on one campaign to another. That was my thought process here.

> I think we probably need to automate this rather than change it to raw_id, like with an admin action that adds all ad network publishers (filter(revshare=70, approved=True)), or RTD publishers (filter(revshare=50)).

I agree something like this is a better long term solution

davidfischer: Just to get some thoughts down, I'm leaning toward a few things:

- Introduce a new model "PublisherGroup". For our use case, we'd probably just have two initially: "Read the Docs Publishers" and "EthicalAds Network Publishers". Campaigns would target these groups and it would be easy to setup a campaign to target one group or multiple groups.
- After that is introduced, remove the `Campaign.publishers` m2m.
- Introduce a new `Campaign.exclude_publishers` m2m where an advertiser can exclude a list of publishers.

davidfischer: Immediately for this PR, I think we remove `publishers` from this raw_id_field so that we can still easily select all publishers.

ericholscher: @davidfischer that sounds good to me. 👍

PR #203, opened 2020-07-28
Advertiser flight improvements

davidfischer:
- Default flight reporting to the start date of the flight. Clicking on a flight and seeing no results is awkward.
- Prompt users to create an ad if there aren't any.
- Break flights into current, past, and upcoming

Reviewing [without whitespace](https://github.com/readthedocs/ethical-ad-server/pull/203/files?w=1) may be helpful.

## Screenshot

Screen Shot 2020-07-28 at 11 55 02 AM

### DIFF
--- a/adserver/templates/adserver/reports/advertiser.html
+++ b/adserver/templates/adserver/reports/advertiser.html
@@ -19,32 +19,38 @@
 <section class="mb-5">
   {% if flights %}
     <h2>{% trans 'Flight report details' %}</h2>
-    <div class="table-responsive">
-      <table class="table table-hover">
-        <thead>
-          <tr>
-            <th><strong>{% trans 'Flight Name' %}</strong></th>
-            <th><strong>{% trans 'Start date' %}</strong></th>
-            <th><strong>{% trans 'End date' %}</strong></th>
-          </tr>
-        </thead>
-        <tbody>
-          {% for flight in flights %}
-            <tr>
-              <td>
-                <a href="{% url 'flight_report' advertiser.slug flight.slug %}?start_date={{ start_date|date:'Y-m-d' }}{% if end_date %}&amp;end_date={{ end_date|date:'Y-m-d' }}{% endif %}">{{ flight.name }}</a>
-                {% if not flight.live %}
-                  <span class="fa fa-eye-slash fa-fw text-muted" aria-hidden="true" data-toggle="tooltip" title="{% trans 'This flight is disabled' %}"></span>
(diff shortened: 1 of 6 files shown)


adserver/templates/adserver/reports/advertiser.html
```diff
@@ -19,32 +19,38 @@
...
-  {% endif %}
+
+    {% regroup flights by state as flight_groups %}
+    {% for state, flight_list in flight_groups %}
+      {% if flight_list %}
+        <div class="table-responsive">
+          <table class="table table-hover">
+            <thead>
+              <tr>
+                <th><strong>{% blocktrans %}{{ state }} flights{% endblocktrans %}</strong></th>
```
ericholscher, line 30: Are we losing the flight name display, or is it shown somewhere else?

adserver/views.py
```diff
@@ -137,24 +134,11 @@ class FlightListView(AdvertiserAccessMixin, UserPassesTestMixin, ListView):
...
-        try:
-            flight_list = paginator.page(self.request.GET.get("page"))
-        except PageNotAnInteger:
-            # If page is not an integer, deliver first page.
-            flight_list = paginator.page(1)
-        except EmptyPage:
-            # If page is out of range (e.g. 9999), deliver last page of results.
-            flight_list = paginator.page(paginator.num_pages)
-
-        context.update({"advertiser": self.advertiser, "flight_list": flight_list})
```
ericholscher, line 157: Is there a reason we're removing pagination? Probably makes sense given there aren't that many flights, just curious.

davidfischer: I removed pagination because we're breaking the flights up into 3 groups. I guess we could still paginate them but does that make sense in the context of groups?

adserver/templates/adserver/reports/advertiser.html
davidfischer: I'm not sure exactly what you mean by that, but the screenshot should illustrate what it will look like. There won't be a table heading "Flight name" but will instead say "Current flights", "Past flights", etc.

adserver/views.py
ericholscher: Yea, I think this is fine, we don't have anyone with hundreds of flights.

adserver/templates/adserver/reports/advertiser.html
ericholscher: Ah yea, I see how it works now. Makes sense 👍

PR #210, opened 2020-07-29
IP Geolocation and Proxy detection improvements

davidfischer:
- Build the basics of an IP deny list based on known proxies
- Adds a Makefile command to update GeoIP databases and proxy lists
- This is just a start and we can deny whole IP ranges in the future


Makefile
```diff
@@ -1,8 +1,14 @@
...
+.PHONY: help test clean dockerbuild dockerserve dockershell dockerprod geoip ipproxy
+
 
-GEOIP_UPDATE = geoipupdate
 GEOIP_DIR = geoip
-GEOIP_CONF_FILE = $(GEOIP_DIR)/GeoIP.conf
+GEOIP_CITY_FILE = dbip-city-lite-2020-07.mmdb.gz
+GEOIP_COUNTRY_FILE = dbip-country-lite-2020-07.mmdb.gz
+GEOIP_CITY_DB_URL = https://download.db-ip.com/free/dbip-city-lite-2020-07.mmdb.gz
+GEOIP_COUNTRY_DB_URL = https://download.db-ip.com/free/dbip-country-lite-2020-07.mmdb.gz
```
humitos, line 8: Looks like the `_FILE` variable should be used to build the URL

Makefile
```diff
@@ -1,8 +1,14 @@
...
+
+TOR_EXIT_LIST_FILE = torbulkexitlist
+TOR_EXIT_LIST_URL = https://check.torproject.org/torbulkexitlist
```
humitos, line 11: Same here.

adserver/tests/test_api.py
```diff
@@ -1079,6 +1079,17 @@ def test_view_tracking_blocked_referrer(self):
...
+    def test_view_tracking_blocked_ip(self):
+        adserver_utils.BLOCKLISTED_IPS = set([self.ip_address])
+
+        resp = self.client.get(self.url)
+
+        self.assertEqual(resp.status_code, 200)
+        self.assertEqual(resp["X-Adserver-Reason"], "Blocked IP impression")
+
+        # Reset the IP blocklist
+        adserver_utils.BLOCKLISTED_IPS = []
```
humitos, line 1091: This looks like a good candidate to be added in a `tearDown` method.

adserver/utils.py
```diff
@@ -216,6 +225,21 @@ def get_geolocation(ip_address):
     return None
 
 
+def build_blocked_ip_set():
+    """Build a set of blocked IPs for preventing bogus ad impressions."""
+    blocked_ips = set()
+
+    filepath = os.path.join(settings.GEOIP_PATH, "torbulkexitlist")
```
humitos, line 232: This file could be a setting. Also, it could be a setting with a list of filenames so we can extend later.

davidfischer: Currently, I think this is unlikely to be extended in that way especially since that would require changes to the `Makefile` and other sources may not come in the same format. Long term, I think we will buy a downloadable database for this.

Makefile
davidfischer: The URLs will probably change while the filename probably won't since it needs to get picked up by the server. I'll make this more clear.

Makefile
davidfischer: I actually want these to be different so the downloadable name can be different from the URL path. The local file name needs to be under our control while the remote file isn't under our control.

adserver/views.py
```diff
@@ -375,6 +376,13 @@ def ignore_tracking_reason(self, request, advertisement, nonce, publisher):
                 user_agent,
             )
             reason = "Blocked referrer impression"
+        elif is_blocklisted_ip(ip_address):
+            log.log(
+                self.log_security_level,
+                "Blocked IP impression, Publisher: [%s]",
+                publisher,
+            )
+            reason = "Blocked IP impression"
```
ericholscher, line 385: Should we log the IP here?

PR #212, opened 2020-07-29
Rate limit ad viewing

davidfischer:
This adds configurable rate limiting for viewing ads. By default, this adds a single limit of 3 views per 5 minutes. The data is short term cached in Redis. We currently have plenty of space in Redis (100s of MBs) but it will be worth monitoring after this goes live.


config/settings/production.py
```diff
@@ -131,6 +131,7 @@
 ADSERVER_CLICK_RATELIMITS = env.list(
     "ADSERVER_CLICK_RATELIMITS", default=["1/m", "3/10m", "10/h", "25/d"]
 )
+ADSERVER_VIEW_RATELIMITS = env.list("ADSERVER_VIEW_RATELIMITS", default=["3/5m"])
```
ericholscher, line 134: This feels a bit aggressive, in terms of a real user session it could easily hit more, especially once you take into account users behind a corporate proxy. Especially now that we have more than 3 paid ads, I think something closer to `5/5m` is probably a better default.

PR #217, opened 2020-07-31
Ignore all known users

davidfischer:
Previously we were ignoring staff but with this change we will ignore all known users. I think this is the right move
so advertisers or publishers impressions don't count. It might make the verification step for publishers a little trickier.

### DIFF
--- a/adserver/tests/test_api.py
+++ b/adserver/tests/test_api.py
@@ -1028,12 +1028,18 @@ class TestProxyViews(BaseApiTest):
         self.assertEqual(resp.status_code, 200)
         self.assertEqual(resp["X-Adserver-Reason"], "Internal IP")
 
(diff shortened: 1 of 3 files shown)


davidfischer:
> This should probably also update the docs to explain this a bit more.

I don't want to give too many details but I agree we could say more. There's some details in the docs [here](https://ethical-ad-server.readthedocs.io/en/latest/user-guide/getting-started.html#understanding-the-ad-server-modeling) (under Clicks) but I'll add a few more details.

ericholscher:
>  There's some details in the docs here (under Clicks) but I'll add a few more details.

@davidfischer yea -- I'm thinking more user-facing docs on ethicalads.io -- or we need to be linking those docs somewhere that users can find them.

PR #219, opened 2020-07-31
Set Sentry log level to error

davidfischer:
We are currently sending a lot to Sentry. This will only send errors. Warnings will still be in our regular log analysis.


davidfischer:
We're going to keep sending warnings to Sentry and instead just reduce the log level on a few common errors.

PR #220, opened 2020-08-01
Admin action for processing refunds

davidfischer:
- Refund clicks and views through the admin interface
- Add a utility if this needs to be done on a larger scale
- This credits advertiser flights and makes refunded views and clicks disappear from reporting


adserver/admin.py
```diff
@@ -719,6 +722,25 @@ def has_add_permission(self, request):
         """Clicks and views cannot be added through the admin."""
         return False
 
+    def refund_impressions(self, request, queryset):
```
ericholscher, line 725: I worry a little bit about not having a confirmation step here, but it should be fine if we accidentally hit it for 1 view or something. But likely need to be a bit more defensive here in the future.

davidfischer: Is there a good way to add a confirmation to the admin?

davidfischer:
> That seems a bit tricky going forward if there's fraud that isn't easy to filter by our existing admin filters, but it seems like a good start at least for this case.

I think for cases where it's hard to filter or where there's a lot, we'll use this from the command line. However, the underlying methods are there so we can use the same code.

adserver/admin.py
davidfischer: There's this: https://docs.djangoproject.com/en/2.2/ref/contrib/admin/actions/#actions-that-provide-intermediate-pages

I'll see if I can add one quickly.

ericholscher:
@davidfischer I think the other option is just to hide the admin action for now -- since it seems unlikely we'll use it much. I imagine most cases of fraud will be systemic and large, requiring the CLI approach.

PR #223, opened 2020-08-04
Save publisher payouts

### LINKED ISSUE #178: Store publisher payout history
We should store publisher payouts and show these to publishers. Each record needs to have the time frame it was for, the amount, how it was paid out, and maybe a link to a confirmation if applicable. Publishers should be able to browse their past payouts.

davidfischer:
Store publisher payout including the date of the payout, the amount, and any confirmation details, notes, or attachments. ~~Because #222 has a migration, this might need an update after that is merged.~~

Fixes #178


adserver/admin.py
```diff
@@ -774,7 +775,17 @@ class ViewAdmin(AdBaseAdmin):
     model = View
 
 
+class PublisherPayoutAdmin(admin.ModelAdmin):
+    list_display = ("amount", "publisher", "date", "modified", "created")
```
ericholscher, line 779: Might show the publisher first here? 
```suggestion
    list_display = ("publisher", "amount", "date", "modified", "created")
```

adserver/templates/adserver/publisher/payouts.html
```diff
@@ -0,0 +1,52 @@
+{% extends "adserver/base.html" %}
```
ericholscher, line 1: Maybe call this template `payout_list.html`? I got confused with just the plural.

adserver/templates/adserver/publisher/payouts.html
```diff
@@ -0,0 +1,52 @@
...
+  {{ block.super }}
+  <li class="breadcrumb-item"><a href="{% url 'publisher_main' publisher.slug %}">{{ publisher }}</a></li>
+  <li class="breadcrumb-item active">{% trans 'Payouts' %}</li>
+{% endblock breadcrumbs %}
+
+
+{% block content_container %}
+
+<h1>{% block heading %}{% trans "Publisher payouts" %}{% endblock heading %}</h1>
+
```
ericholscher, line 19: We should probably have some copy here explaining our payout policies, and links to some docs perhaps? Definitely need at least some introductory copy.

adserver/views.py
```diff
@@ -823,6 +825,55 @@ def get_success_url(self):
...
+
+    model = PublisherPayout
+    template_name = "adserver/publisher/payouts.html"
+
+    def get_context_data(self, **kwargs):  # pylint: disable=arguments-differ
+        context = super().get_context_data(**kwargs)
+
+        payouts = self.get_queryset()
+        total = payouts.aggregate(
+            total=models.Sum("amount", output_field=models.FloatField())
```
ericholscher, line 840: Seems odd to use a Float to count money, is there not Decimal support here?

adserver/admin.py
davidfischer: My reasoning for not doing that is that makes the publisher name the clickable link but it leads to a payout, not to the publisher. I also considered maybe displaying the PK as the clickable link. Thoughts?

adserver/views.py
davidfischer: The underlying field where the payout is stored is an exact decimal field. Only when it is aggregated is it rounded into a float field. I didn't actually test whether a decimal field was supported here.

adserver/templates/adserver/publisher/payouts.html
davidfischer: Where should those docs live? Not in the ad server, I'd think.

adserver/views.py
ericholscher: Yea -- I think we need to get the website built out. Having a place for blog posts, policy, etc. seems more and more important.

davidfischer: A decimal field will work fine for aggregating.

adserver/admin.py
ericholscher: I think its probably fine as you have it, for now.

davidfischer: I ended up pushing a change that linked the PK.

davidfischer:
I pushed a few changes with some fixes:

- Fix the migration order
- Use decimalfield for aggregating
- payouts -> payout-list
- clickable link for a payout PK

PR #226, opened 2020-08-07
Add EthicalAds branding to the adserver

davidfischer:
- Conditional on the ADSERVER_ETHICALADS_BRANDING envvar
- Default false (false in dev)
- Displays "Ethical Ad Server" if branding is off
- Fixed an issue with toggling navigation on mobile as well


assets/src/img/ethicalads-logo-blue.svg
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ericholscher, line 1: I feel like we shouldn't even be checking these assets into the repo. There should be an env var for a link that contains them. We don't want them licensed under the same license as the code, and we definitely don't want to distribute them, which is the issue we hit with RTD.

assets/src/scss/_branding.scss
```diff
@@ -0,0 +1,31 @@
+// Anyone may use the ad server under the terms of the license.
+// However, permission to use the EthicalAds brand, logo, and trademarks,
+// are not conferred with the permission to use the code.
```
ericholscher, line 3: We should avoid this entirely and just not ship the logos.

davidfischer:
> I'm pretty sure we definitely shouldn't be shipping our logos in this repo. There should be an option to have a remote CSS/Logo file, and we can use that, but giving people the logos and just an option to turn them on isn't any different than what we're doing with RTD.

From a legal perspective, I don't think including them in the repo is an issue. You can license them separately.

As to your other point, I think there are trade-offs that are worth acknowledging. I considered some of these and came to the opposite conclusion, but I'll outline them here so we can discuss them and come to a consensus.

* Including just a logo via an envvar seems insufficient. There might be additional things you'd want separate from just a logo. I think a complete CSS file which might link to additional files like images is what we want. One advantage of having them in the same repo is to use the same SASS variables although that's pretty minor, I'd say.
* Having the files in the repo does simplify deployment quite a bit and testing to a small degree. Having non-versioned files that live in a separate repo and are deployed differently seems suboptimal. It might be possible to have an add-on repo like we have for Read the Docs although that has its own issues.
* I can understand your concern about shipping the logos and maybe "distributing" them means something special in this context. However, there's no issue having a few files in a repo under a separate license and as long as the default is to *not* use them I'm not expecting a bunch of folks to use our branding.

davidfischer:
After discussing, I think we're going to keep the CSS in the repo, move the logos outside of the repo, and make the logo URLs a SCSS variable so other folks can easily override them.

assets/src/scss/_branding.scss
davidfischer: Moved the logos out

PR #229, opened 2020-08-19
Store publisher payout settings

### LINKED ISSUE #177: Store publisher payout settings
For each publisher we should store their payout preferences. To start with, this is probably just how they prefer to be paid out (PayPal, Wire, check, etc.) and any relevant details about them (email address, postal address, etc.).

davidfischer:
- Store open collective name, paypal email, or stripe info
- Allow stripe oauth workflow for connected accounts
- Store payout method on the payout object itself

Fixes #177

## Testing
To fully test this, you'll need to get the Stripe [test key](https://dashboard.stripe.com/test/apikeys) and Stripe [test client ID](https://connect.stripe.com/test/settings/applications) from the Stripe dashboard. You can put them in `.envs/local/django` and start the docker setup. You should be able to configure the publisher payout settings and connect to stripe.


adserver/forms.py
```diff
@@ -228,8 +236,53 @@ def __init__(self, *args, **kwargs):
...
+            stripe_block = HTML("<!-- Stripe is not configured -->")
+
         self.helper = FormHelper()
         self.helper.layout = Layout(
+            Fieldset(
+                _("Payout settings"),
+                Field("payout_method", data_bind="value: payoutMethod"),
+                Div(stripe_block, data_bind="visible: (payoutMethod() == 'stripe')"),
+                Div(
+                    PrependedText(
```
ericholscher, line 275: This is really nice.

adserver/models.py
```diff
@@ -134,6 +135,28 @@ class Publisher(TimeStampedModel, IndestructibleModel):
...
+        null=True,
+        default=None,
+    )
+    open_collective_name = models.CharField(
+        _("Open Collective name"), max_length=200, blank=True, null=True, default=None
+    )
+    paypal_email = models.EmailField(
+        _("PayPal email address"), blank=True, null=True, default=None
+    )
+
```
ericholscher, line 159: Should we have some kind of content block for `Other`? I'm thinking we should probably just not include it at this point, unless it's required for showing `None`?

adserver/views.py
```diff
@@ -818,6 +821,94 @@ def get_success_url(self):
...
+    permanent = False
+
+    def get_redirect_url(self, *args, **kwargs):
+        if not settings.STRIPE_CONNECT_CLIENT_ID:
+            messages.error(self.request, _("Stripe is not configured"))
+            return reverse("dashboard-home")
+
+        publisher = self.get_object()
+
+        stripe_state = get_random_string(30)
```
ericholscher, line 840: Might be useful to explain this a bit. I assume it's a secret we verify at some point?

davidfischer: That's fair. I'll add a note. Yes this is just returned after the stripe oauth connect so we only save values that were redirected from stripe.

adserver/models.py
davidfischer: It isn't required although it is useful especially for showing the payout method on payout objects themselves.

davidfischer:
> I think this could be how we track all publishers $$

For connected accounts, you do pay [$2 per connected account per month](https://stripe.com/connect/pricing) I believe.

ericholscher:
@davidfischer Yea I saw that, but it seems to only count accounts we've sent money to that month:


Screen Shot 2020-08-20 at 9 11 24 AM

davidfischer:
Ahhh, ok. That's good

PR #231, opened 2020-08-20
Use Django2 style URLs everywhere

davidfischer:
This is a minor change to use new style URLs everywhere rather than old style ones. This matches [the Django docs](https://docs.djangoproject.com/en/2.2/topics/http/urls/).


davidfischer:
This has been a really minor thing I wanted to do for a while.

PR #236, opened 2020-09-02
Add a management command to add a publisher

ericholscher:
This will make it a bit quicker to add publishers,
and make sure we don’t miss steps like adding them to the EA group.


davidfischer:
I updated a few things here and it's worth a full review. I added some tests and refactored a bit to clean things up. I didn't want to be creating a fake request in the management command which just felt weird. Instead it uses the default site.

davidfischer (approved):

ericholscher (commented):
Looks good :+1:

PR #237, opened 2020-09-14
Improve automation around payouts

ericholscher:
This is a stopgap until we add some kind of monthly balance display,
since that is the real data that we want for accounting and historical purposes.

Currently we don't track the timeframe that a payout is for,
which would make this code more explicit.
We could then also generate links to the ad report for that timeframe.


adserver/management/commands/payouts.py
```diff
@@ -0,0 +1,147 @@
...
+        for publisher in Publisher.objects.all():
+            # user = publisher.user_set.first()
+            last_payout = publisher.payouts.last()
+            first = False
+
+            if not last_payout:
+                first = True
+                # Fake a payout from 2020-01-01 to make the logic work.
+                fake_date = datetime(year=2020, month=1, day=1)
+                last_payout = PublisherPayout(date=fake_date, amount=0)
```
davidfischer, line 105: This is a little awkward. Since we only use the `last_payout` for the date, maybe just use a date object instead of faking a `PublisherPayout`?

adserver/management/commands/payouts.py
```diff
@@ -0,0 +1,147 @@
...
+                    )
+                    + "?start_date="
+                    + last_payout.date.strftime("%Y-%m-01")
+                    + "&end_date="
+                    + month.strftime("%Y-%m-%d")
+                )
+                print(publisher, due_str)
+                print("CTR {:.3f}".format(ctr))
+                print("https://server.ethicalads.io" + report_url)
+                print(
```
davidfischer, line 126:
```suggestion
                self.stdout.write(f"{publisher} {due_str}")
                self.stdout.write("CTR {:.3f}".format(ctr))
                self.stdout.write("https://server.ethicalads.io" + report_url)
                self.stdout.write(
```

Rather than `print`, use `self.stdout.write`. This makes it easier to test and capture the stdout.

adserver/models.py
```diff
@@ -1567,7 +1602,8 @@ class PublisherPayout(TimeStampedModel):
     )
 
     class Meta:
-        ordering = ("-date",)
+        # This is 'date' instead of '-date' to make `first()` and `last()` work properly
+        ordering = ("date",)
```
davidfischer, line 1606: This requires a migration (why is a good question, but it does) and it will order the oldest payouts at the top of the page on the payouts screen. Do you think that's a little strange? I do realize that using `.last()` and it gets the first payout is a bit awkward too.

davidfischer (commented):
Copy looks good although I left a few changes.

adserver/models.py
ericholscher: I think we can just force the ordering on the page display if we want. The models should definitely default to time-based ordering I think.

adserver/management/commands/payouts.py
ericholscher: Yea, that probably makes sense.

adserver/models.py
ericholscher: I pushed up this change.

adserver/utils.py
```diff
@@ -319,6 +321,73 @@ def generate_absolute_url(view, kwargs):
...
+    last_payout = publisher.payouts.last()
+
+    if last_payout:
+        first = False
+        # First of the month of the month the payout was for.
+        # TODO: Store this data on the model, instead of hacking it.
+        last_payout_date = last_payout.date.replace(day=1)
+    else:
+        first = True
+        # Fake a payout from 2020-07-01 to make the logic work.
```
davidfischer, line 337:
```suggestion
        # Fake a payout to make the logic work.
```

davidfischer (approved):
I could see this table getting a bit awkward once a publisher has a lot of payouts. Maybe their pending balance and "this month" should be outside of the table at the top under the description? Nonetheless, this is definitely better than what we have now.

Screen Shot 2020-09-16 at 3 28 02 PM

ericholscher:
@davidfischer yea -- I'd like to pull this data into the top-level default for publishers, but at least for now we're calculating and displaying it 👍

PR #239, opened 2020-09-18
Store placements and keywords and add reporting

ericholscher:
This adds an initial Placements report,
it is behind a feature flag for staff currently.
It also starts storing data on placements and keywords,
so that we can start to aggregate over time.

Depends on  https://github.com/readthedocs/ethical-ad-client/pull/30


davidfischer (commented):
How will this work with the current system? The current ad client sends the div ID as just `ad_<timestamp>` so won't each one be unique?

There seems to be a bit of extra work here since we aren't storing a record for offering an ad. Maybe we should just store an ad offer. The offer would contain the keywords for which it was offered and other metadata. I've wanted something similar in the past for other reasons (auditing and fraud prevention).

adserver/models.py
```diff
@@ -1437,6 +1476,36 @@ def __str__(self):
...
 
+class PlacementImpression(BaseImpression):
+
+    """
+    Create an index of placements for ads.
+
+    Indexed one per ad per day per publisher.
+    """
+
+    placement = models.CharField(max_length=255, null=True, blank=True)
```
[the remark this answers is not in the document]
davidfischer: The ad type is worth capturing since we could then find out things like "this ad type has a low view to offers ratio".

adserver/views.py
```diff
@@ -768,6 +776,42 @@ def get_context_data(self, **kwargs):
...
+        report = publisher.daily_reports(
+            start_date=context["start_date"],
+            end_date=context["end_date"],
+            campaign_type=context["campaign_type"],
+            placement=context["placement"],
+        )
+
+        placement_options = publisher.placement_impressions.values_list(
+            "placement", flat=True
+        ).distinct()[: self.PLACEMENT_LIMIT]
```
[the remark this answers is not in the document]
ericholscher: Yea, I think we want to limit them on creation perhaps, but I haven't thought exactly about how we want to handle it.

adserver/models.py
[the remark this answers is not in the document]
ericholscher: Cool, I can add that to the index 👍

adserver/views.py
```diff
@@ -414,6 +414,7 @@ def get(self, request, advertisement_id, nonce):
         """Handles proxying ad views and clicks and collecting metrics on them."""
         advertisement = get_object_or_404(Advertisement, pk=advertisement_id)
         publisher = advertisement.get_publisher(nonce)
+        div_id, keywords, ad_type = advertisement.get_request_data(nonce)
```
ericholscher, line 417: Wondering if it makes sense to just use this as `request_data` and pass it along to each of the functions as that single argument? Might be cleaner.

adserver/models.py
```diff
@@ -1242,50 +1227,20 @@ def offer_ad(self, publisher, ad_type_slug, div_id=None, keywords=None):
...
         """
-        return (
-            cache.get(
-                self.cache_key(impression_type=impression_type, nonce=nonce), None
-            )
-            == 0
-        )
+        four_hours_ago = timezone.now() - datetime.timedelta(hours=4)
+        return Offer.objects.filter(
+            id=nonce, date__lte=four_hours_ago, viewed=False
```
davidfischer, line 1239: This should be greater than, right?

davidfischer: Hopefully postgres is smart enough to optimize this query correctly. I wonder if just fetching and comparing would be better.

ericholscher: Yea, I ended up refactoring this for a couple reasons and it should be clearer now.

adserver/templates/adserver/base.html
```diff
@@ -90,9 +90,19 @@ <h6 class="text-muted">{{ publisher }}</h6>
...
+                  <span class="fa fa-bar-chart fa-fw ml-4 text-muted" aria-hidden="true"></span>
+                  <span>{% trans 'Placements' %}</span>
+                </a>
+              </li>
+              {% endif %}
+
               <li class="nav-item">
                 <a class="nav-link" href="{% url 'publisher_payouts' publisher.slug %}">
-                  <span class="fa fa-dollar fa-fw mr-2 text-muted" aria-hidden="true"></span>
+                  <span class="fa fa-dollar fa-fw mr-2 text-muted" aria-hidden="true">-</span>
```
davidfischer, line 105:
```suggestion
                  <span class="fa fa-dollar fa-fw mr-2 text-muted" aria-hidden="true"></span>
```

adserver/templates/adserver/base.html
```diff
@@ -90,9 +90,19 @@ <h6 class="text-muted">{{ publisher }}</h6>
                 </a>
               </li>
 
+              {% if request.user.is_staff %}
```
davidfischer, line 93:
```suggestion
              {% if request.user.is_staff and publisher.record_placements %}
```

Since the `record_placements` is always present in the settings form, should we just make this available if the publisher has it turned on?

adserver/views.py
```diff
@@ -413,22 +415,24 @@ def send_to_analytics(self, request, advertisement, message):
     def get(self, request, advertisement_id, nonce):
         """Handles proxying ad views and clicks and collecting metrics on them."""
         advertisement = get_object_or_404(Advertisement, pk=advertisement_id)
-        publisher = advertisement.get_publisher(nonce)
+        offer = Offer.objects.filter(id=nonce).first()
```
davidfischer, line 418: This can be none. In that case, we won't know the publisher and the view shouldn't count (somebody just made up a nonce/offer). This function as written will 500 since it depends on `offer.publisher`.

adserver/views.py
```diff
@@ -768,6 +774,43 @@ def get_context_data(self, **kwargs):
         return context
 
 
+class PublisherPlacementReportView(PublisherAccessMixin, BaseReportView):
```
davidfischer, line 777: I do wonder if we even need this report at all. The only difference is the ability to filter by div. Couldn't we just only show the Divs dropdown on the regular publisher report if the publisher has the flag turned on?

adserver/admin.py
```diff
@@ -770,6 +780,13 @@ def refund_impressions(self, request, queryset):
         return None
 
 
+class OfferAdmin(AdBaseAdmin):
+
+    """Django admin configuration for ad offers."""
+
+    model = Offer
```
davidfischer, line 787: Should we add `viewed` and `clicked` here? Might be useful in dev but probably not very useful in prod.

davidfischer (commented):
This looks much better than it was! I like relying on the DB instead of redis for something ephemeral and I think this will give us more details.

I did have a few bits of feedback. Most of them were minor but I was able to trigger a 500.

adserver/templates/adserver/base.html
ericholscher: Yea, is_staff is leftover from the old PR that didn't have the option, we can remove it.

adserver/views.py
ericholscher: Yea, I had it as a `get_object_or_404` but it was raising some odd validation errors from Django when passing invalid nonce's. It should return from the `ignore_tracking_reason` when the nonce is invalid, so I will just fix the logic here to let `publisher` be `None`, which should be accounted for.

adserver/views.py
ericholscher: Yea, that's how I had it originally, but since it's using a different table, "All placements" will show a different report than "All views" because we aren't always recording a placement. We could have one report with a default of None, and then "all" be a different view, but it feels weird.

adserver/admin.py
ericholscher: Makes sense :+1:

adserver/views.py
davidfischer: I guess also we can't backfill it.

adserver/views.py
davidfischer: It still throws a 500 if for a URL like `/proxy/view/698/blah/` since `blah` is an invalid UUID. This will cause an issue with all the existing nonces out there when we roll it out (anyone clicking those will trigger a 500). It isn't a huge deal but we can avoid it. I think ideally, it should just not count that for billing and redirect appropriately.

davidfischer (approved):
This looks good. We should be careful when we roll it out as it does significantly change. However, I think we should be good.

I did note one UUID/nonce error.

adserver/views.py
ericholscher: Yea, if you have thoughts I'm happy to hear them. Should we be catching that exception and just setting `offer` to `None`?

ericholscher:
@davidfischer one last review and I think it's ready to go!

adserver/views.py
ericholscher: I've added this, and it no longer 500's.

davidfischer (approved):
I reviewed the changes since my last approval and everything looks good.

The only question I have is that since the `div_id` sent by the existing ad client is `ad_<date>` and that will always be unique, should we ignore it?

ericholscher:
We already are ignoring it: https://github.com/readthedocs/ethical-ad-server/pull/239/files#diff-36aca7ee8bd2b9b7cae23905d47d8b9fR1067

ericholscher:
Gonna ship this! 🎉

PR #243, opened 2020-09-29
Add a simple “Advertiser” filter to the publisher templates

ericholscher:
This is just a stopgap until we have a proper way to show more detailed per-advertiser data in the reports.
I just didn’t want to change up the reporting logic quite yet,
and this was easy enough to get the data we need.

### DIFF
--- a/adserver/views.py
+++ b/adserver/views.py
@@ -1006,6 +1006,15 @@ class AllPublisherReportView(BaseReportView):
 
         publishers = Publisher.objects.filter(id__in=impressions.values("publisher"))
 
+        advertiser_list = (
+            impressions.order_by("advertisement__flight__campaign__advertiser__slug")
+            .values_list(
+                "advertisement__flight__campaign__advertiser__slug",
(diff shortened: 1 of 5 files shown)


ericholscher:
We should probably add this to the All publishers report as well, now that I think about it.

davidfischer (approved):
We were going to limit the possible advertiser choices on the publisher report but this looks good.

ericholscher:
@davidfischer alright updated the query 👍

adserver/views.py
```diff
@@ -759,18 +760,27 @@ def get_context_data(self, **kwargs):
...
+            .order_by("-total_views")
+            .filter(total_views__gt=100)
             .values_list(
                 "advertisement__flight__campaign__advertiser__slug",
                 "advertisement__flight__campaign__advertiser__name",
             )
-            .distinct()
         )
 
+        if context["start_date"]:
```
davidfischer, line 775: Shouldn't we do these filters before aggregating?

ericholscher:
Shipping!

PR #244, opened 2020-09-30
Add an initial Geo report for publishers

ericholscher:
This builds off our existing model with Placements, and provides filtering by geo. We will eventually do a breakdown of geos in the reporting, but this at least captures the data.

This also adds some additional context to each of the reports, so users know what they are. 

In order to test it, I had to hack the `get_client_country` function to just return a fake country locally. That at least allowed GeoImpressions to get written.


adserver/migrations/0034_add_geo_placement.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('adserver', '0033_add_keyword_placement_impressions'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
```
davidfischer, line 16: I believe this migration probably needs to be recreated after merging master. I think you took an older version of the placement migration and that's why we're failing the migrations check.

adserver/models.py
```diff
@@ -1073,6 +1082,17 @@ def incr(self, impression_type, publisher, div_id=None, ad_type_slug=None):
                 **{impression_type: models.F(impression_type) + 1}
             )
 
+        if publisher.record_geos and request:
```
davidfischer, line 1087: Taken by itself, I'm not too concerned here. However, the long term solution is to rollup this data every few hours or every day. I'd like to do as little as possible in the request to keep the performance fast and to keep it simple. I do wonder whether this should just wait for the infrastructure changes needed to do that. I feel like we're writing code we're going to throw away (models wouldn't go away but this section would).

davidfischer:
I fixed up the linting problem so that just leaves the migration problem, I think.

ericholscher:
@davidfischer if we're not scared about the DB issue from last week, I'd still like to get this shipped this week.

adserver/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+from .models import GeoImpression
+from .models import Offer
+from .utils import get_ad_day
+from config.celery_app import app
+
+log = logging.getLogger(__name__)  # noqa
+
+
+@app.task()
+def daily_update_geos():
```
ericholscher, line 15: This should probably eventually take a `day` or `date_range` argument, but just wanted to keep it simple for now.

davidfischer: It definitely should since without this we won't be able to run it on past data.

adserver/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+    """Update the Geo index each day."""
+    beginning_of_today = get_ad_day()
+    log.info("Updating GeoImpressions for %s", beginning_of_today)
+
+    for values in (
+        Offer.objects.filter(date__gte=beginning_of_today, viewed=True)
```
davidfischer, line 21: I would also constrain on `lt=beginning_of_day + timedelta(days=1)` or something like that. There could be some timing issues with this code right around midnight UTC.

adserver/templates/adserver/base.html
```diff
@@ -99,6 +99,13 @@ <h6 class="text-muted">{{ publisher }}</h6>
               </li>
               {% endif %}
 
+              <li class="nav-item">
```
davidfischer, line 102: Since there won't immediately be data here and we don't yet have a way to backfill this data, should we hide this report during a rollout process?

adserver/templates/adserver/base.html
```diff
@@ -99,6 +99,13 @@ <h6 class="text-muted">{{ publisher }}</h6>
...
+                <a class="nav-link" href="{% url 'publisher_geo_report' publisher.slug %}">
+                  <span class="fa fa-bar-chart fa-fw ml-4 text-muted" aria-hidden="true"></span>
```
davidfischer, line 104:
```suggestion
                  <span class="fa fa-globe fa-fw ml-4 text-muted" aria-hidden="true"></span>
```

config/settings/production.py
```diff
@@ -106,6 +108,12 @@
 CELERY_RESULT_BACKEND = CELERY_BROKER_URL
 CELERY_BROKER_USE_SSL = {"ssl_cert_reqs": ssl.CERT_REQUIRED}
 
+CELERYBEAT_SCHEDULE = {
+    "every-day-generate-geo-index": {
+        "task": "adserver.tasks.daily_update_geos",
+        "schedule": crontab(minute="*/30"),
```
davidfischer, line 114: Should we actually make this a daily task? Or are you thinking this during the rollout process?

adserver/tasks.py
```diff
@@ -0,0 +1,52 @@
...
+            publisher_id=values["publisher"],
+            advertisement_id=values["advertisement"],
+            country=values["country"],
+            date=beginning_of_today,
+        )
+        GeoImpression.objects.filter(pk=impression.pk).update(
+            views=values["country__count"]
+        )
+
+    for values in (
```
davidfischer, line 37: We have blocks to get the views and the clicks, but not one to get the offers for a geo. I think we should add that so we can compare view/click to offer ratios for various countries.

davidfischer (approved):
This tested out clean and worked well in development. This is really nice and I like this pattern a lot. I had a few minor questions and bits of feedback but this is ready to go.

config/settings/production.py
ericholscher: I think it should probably be daily, but wanted it faster for the rollout, yea.

adserver/tasks.py
ericholscher: Gotcha, I will add that as well 👍

adserver/templates/adserver/base.html
ericholscher: I've made it staff protected for now 👍

davidfischer:
This tested out well! Let's go live with it.

PR #250, opened 2020-10-08
Add total breakdown to All publishers report

ericholscher:
This will enable us to more easily compare Day over Day metrics across a subset of our stack.

### DIFF
--- a/adserver/templates/adserver/reports/all-publishers.html
+++ b/adserver/templates/adserver/reports/all-publishers.html
@@ -70,6 +70,19 @@
           </tr>
         </thead>
         <tbody>
+          {% for date, day in days.items %}
+            {% if day.views > 0 or day.clicks > 0 %}
+            <tr>
+              <td>{{ date }}</td>
+              <td class="text-right">{{ day.views|intcomma }}</td>
+              <td class="text-right">{{ day.clicks|intcomma }}</td>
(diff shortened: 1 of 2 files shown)


adserver/templates/adserver/reports/all-publishers.html
```diff
@@ -40,12 +40,14 @@
 {% block toc %}
 <section class="mb-5">
   {% if publishers %}
-    <h2>{% trans 'Publishers '%}</h2>
+    <h2>{% trans 'Publishers '%} ({{ publishers|length }})</h2>
+    <details>
```
davidfischer, line 44: I assume you're aware but this will hide these by default. I think I'm OK with that.

davidfischer (approved):
This is great and I wanted this again today!

adserver/templates/adserver/reports/all-publishers.html
ericholscher: Yea -- I don't think its super valuable to take up the first 2 pages of screen, and easy enough to open if needed.

ericholscher: @davidfischer I actually want a way to hide the per-publisher breakdown table as well, but `details` doesn't work on tables, so I punted for now :)

PR #257, opened 2020-10-19
Index the date fields on ad impressions, clicks, views, and offers

davidfischer:
This migration indexes a few potentially *VERY LARGE* tables.
As a result, it is probably best faked and run concurrently.

https://www.postgresql.org/docs/9.1/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY


ericholscher (approved):
Looks good. We should make sure to fake this prior to deploy 👍

davidfischer:
> Looks good. We should make sure to fake this prior to deploy

The indexes are applied.

PR #260, opened 2020-10-22
Migrate PlacementImpressions to a Celery task

ericholscher:
This moves our only other in process index into a celery task. We now only
update the ``Offer`` and ``AdImrpession`` table on the AdDecision api.

### DIFF
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -1,6 +1,7 @@
 """Celery tasks for the ad server."""
 import datetime
 import logging
+import re
 
 from django.db.models import Count
 
@@ -9,6 +10,7 @@ from .constants import IMPRESSION_TYPES
 from .constants import VIEWS
 from .models import GeoImpression
 from .models import Offer
+from .models import PlacementImpression
 from .utils import get_ad_day
 from config.celery_app import app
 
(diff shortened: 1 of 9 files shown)


adserver/tasks.py
```diff
@@ -55,3 +56,49 @@ def daily_update_geos(day=None):
...
+@app.task()
+def daily_update_placements(day=None):
+    """
+    Update the Placement index each day.
+
+    :arg day: An optional datetime object representing a day
+    """
+    start_date = get_ad_day()
+    if day:
+        start_date = day.replace(hour=0, minute=0, second=0, microsecond=0)
```
davidfischer, line 70: This looks fine but does mean we're sending a complex object (datetime+timezone) through celery which might not be ideal. I think we had some issues related to this on RTD.

davidfischer (approved):
This looks good.

One issue with rolling this out in production is that we have offers dating back to a specific date but before that we don't so we have to be run this task only up until dates where we have data. Otherwise, we'll create aggregated records with 0 for clicks/views/offers.

ericholscher:
I don't plan to backfill this data, since it was already recording. It will just start on the day we ship it.

adserver/tasks.py
ericholscher: Hrm yea -- we should probably just be passing in a string then calling `dateutil.parse` on it.

ericholscher: Should be working now, tested it locally 👍

PR #262, opened 2020-10-26
Remove unused entrypoint from dockerfile

ericholscher:
This seems like some kind of cruft from the initial layout,
but it isn’t used and makes it harder to set a custom DATABASE_URL


davidfischer:
I think this is what gets used when you just open a docker shell. Maybe I'm mistaken?

ericholscher:
I don't believe it's really used, except to wait for Postgres to come up, but that isn't really necessary, and breaks when you set the `DATABASE_URL` directly, which is annoying.

davidfischer (approved):
I ended up testing it through and it seems ok. There's probably some problem of not waiting for the DB to be up but it probably won't manifest itself in practice. It also won't affect production since this isn't used there.

PR #263, opened 2020-10-27
Show breakdown report on the Geo/Placement reports by default

ericholscher:
This still shows day-by-day data when filtered,
but defaults to a breakdown of the top 20 items.


davidfischer (approved):
This tested out pretty nicely. These seem much more useful.

ericholscher:
Yea, I think we're getting close to shipping them by default, and then adding more 🎉

PR #265, opened 2020-10-27
Add an estimated count paginator

davidfischer:
We have a few very slow tables that won't load in the admin without this.

### DIFF
--- a/adserver/mixins.py
+++ b/adserver/mixins.py
@@ -59,7 +59,10 @@ class EstimatedCountPaginator(Paginator):
 
     @cached_property
     def count(self):
-        if "postgresql" not in settings.DATABASES["default"]["ENGINE"]:
+        if (
+            settings.DEBUG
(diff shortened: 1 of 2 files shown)


adserver/mixins.py
```diff
@@ -43,3 +47,29 @@ def test_func(self):
...
+
+    """
+    Paginator that gives only an estimated count based on DB statistics..
+
+    This only works for PostgreSQL. Other database engines always return the full count.
+    """
+
+    @cached_property
+    def count(self):
+        if "postgresql" not in settings.DATABASES["default"]["ENGINE"]:
```
ericholscher, line 62: Could turn it off when `DEBUG` is `True`, but doesn't seem too necessary.

ericholscher (approved):
Seems simple enough 👍

adserver/mixins.py
davidfischer: That's a good idea!

PR #268, opened 2020-10-27
Add a date hierarchy to the impression admin views

davidfischer:


ericholscher:
This seems to perform badly in prod from my testing. Appears to hang up on `SELECT DISTINCT DATE_TRUNC('day', "adserver_offer"."date") AS "datefield" FROM
 "adserver_offer" WHERE "adserver_offer"."date" IS NOT NULL ORDER BY "datefield" ASC`

ericholscher (changes requested):
Unless you found different, this seemed to break the admin pages locally when I tried it. :/

davidfischer:
I think you're right. There's possibly some options here but maybe nothing out of the box. I'm going to give it a quick look and close this out if I don't find a good option.

davidfischer:
Here's [an article](https://hakibenita.medium.com/scaling-django-admin-date-hierarchy-85c8e441dd4c) outlining some of the trade-offs. I'm fine closing this out for now.

PR #269, opened 2020-10-28
Add a render_pixel option to the publisher.

ericholscher:
This allows us to transition to viewport tracking:

* Publisher gets send_pixel field (default True)
* Ad Type checks {% if send_pixel %} around the pixel
* We ship the client with viewport tracking
* We disable send_pixel on the users we want viewport tracking on


adserver/models.py
```diff
@@ -175,6 +175,12 @@ class Publisher(TimeStampedModel, IndestructibleModel):
     record_placements = models.BooleanField(
         default=False, help_text=_("Record placement impressions for this publisher")
     )
+    render_pixel = models.BooleanField(
+        default=True,
```
ericholscher, line 179: Actually, I wonder if this should default False, so all new publishers need to use the client, and we can enable it if they want a custom integration. 

I think for now we default True for the migration, but eventually switch to default False

adserver/templates/adserver/advertisement.html
```diff
@@ -6,7 +6,7 @@
     <a href="{{ link_url }}" rel="nofollow" target="_blank">
       <img src="{{ image_url }}">
     </a>
-  {% elif view_url %}
+  {% elif view_url and publisher.render_pixel %}
```
davidfischer, line 9: This is only the default template that isn't used. The real one is stored in the DB.

davidfischer (approved):
This looks good although it might be worth it to have a test.

adserver/templates/adserver/advertisement.html
ericholscher: Yea, I know -- was just an example and to keep things consistent.

ericholscher:
Weird, that test passes for me locally -- I'm not convinced it's a useful test tho :/

davidfischer:
> I'm not convinced it's a useful test tho :/

I think it's fine. :shipit:

PR #271, opened 2020-10-29
Add a new report for Publishers showing their advertisers

ericholscher:
This allows publishers to understand who is running ads on their site,
and the relative performance between them.

This is currently in beta and the menu is only shown to staff.

### DIFF
--- a/adserver/models.py
+++ b/adserver/models.py
@@ -3,6 +3,7 @@ import datetime
 import html
 import logging
 import math
+import operator
 import uuid
 from collections import Counter
 from collections import defaultdict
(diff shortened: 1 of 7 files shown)


davidfischer (approved):
This looks good. I believe publishers can already pick a single advertiser and see the results filtered. This will remove that in favor of the advertiser report. I don't know if many publishers look at it but we should probably roll out the wider report (which is staff-only with this PR) to everyone soonish so they don't feel like they lost functionality.

ericholscher:
Yea -- you can still filter the advertiser report by advertiser, for the same view. I'd be fine shipping this to everyone if you'd prefer.

davidfischer:
We can roll it out for a week and then do that. We just don't want to keep people waiting.

PR #273, opened 2020-11-02
Add ability to sort All Publishers report by all metrics

ericholscher:
This gives us more flexibility in looking at reports.

### DIFF
--- a/adserver/views.py
+++ b/adserver/views.py
@@ -1177,12 +1177,16 @@ class AllPublisherReportView(BaseReportView):
                 publishers_and_reports.append((publisher, report))
 
         # Sort reports by revenue
-        if sort == "revenue":
-            publishers_and_reports = sorted(
-                publishers_and_reports,
-                key=lambda obj: obj[1]["total"]["revenue"],
-                reverse=True,
-            )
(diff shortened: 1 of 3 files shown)


adserver/views.py
```diff
@@ -1177,10 +1177,11 @@ def get_context_data(self, **kwargs):  # pylint: disable=too-many-locals
                 publishers_and_reports.append((publisher, report))
 
         # Sort reports by revenue
-        if sort == "revenue":
+        sort_options = report["total"].keys()
```
davidfischer, line 1180: By passing this directly to the display, it does result in slightly awkward display but it is just for us so I guess it's ok. I'm not sure every field is useful either.

Screen Shot 2020-11-03 at 11 35 39 AM

davidfischer (approved):
I think the display is slightly awkward but it is staff only and could be refined later. Overall, this worked well in testing.

adserver/views.py
ericholscher: Yea, it's ugly for was easy enough. I'd also like more complex filtering (eg. views__gt=500) or similar, but baby steps :)

PR #274, opened 2020-11-02
Add Decision modeling to our indexes

ericholscher:
This gives us reporting around Fill Rate.
Currently this is an admin-only view.
I thought about having a Fill Rate report,
but realized it’s a stat we probably want for Geo and Placement,
along with against all our traffic.


adserver/models.py
```diff
@@ -1476,6 +1476,18 @@ class BaseImpression(TimeStampedModel, models.Model):
 
     date = models.DateField(_("Date"), db_index=True)
 
+    # Decisions are a superset of all Offers.
+    # Every API request that comes in results in a Decision,
+    # and an Offer is only created when we actually offer an ad.
+    decisions = models.PositiveIntegerField(
```
ericholscher, line 1482: I'm more worried about this migration, since it adds a field to all our indexes. Hopefully it should be fast tho, and only impact reporting.

adserver/models.py
```diff
@@ -1106,15 +1127,20 @@ def get_absolute_url(self):
...
-    def incr(
-        self, impression_type, publisher, request=None, div_id=None, ad_type_slug=None
-    ):
-        """Add to the number of times this action has been performed, stored in the DB."""
+    def incr(self, impression_type, publisher):
+        """
+        Add to the number of times this action has been performed, stored in the DB.
+
+        TODO: Refactor this method, moving it off the Advertisement class since it can be called
+              without an advertisement when we have a Decision and no Offer.
```
davidfischer, line 1135: This is pretty awkward. I wasn't even sure calling this method without an instance was valid. Would making this a `classmethod` be better?

davidfischer (commented):
I don't think the migrations will be that bad. I do think the call to `incr` is pretty awkward. I wasn't even sure it would work at first and that makes me think we should probably do something different.

adserver/models.py
ericholscher: Probably. I think probably moving it to the Publisher is more correct, since that is never null. Didn't seem like a huge priority right now tho, and would have been even more refactoring in the PR.

ericholscher: I tried to refactor it, and it still leads to some janky logic like `publisher.incr(impression_type=CLICKS, advertisement=self)` -- doesn't seem much better.

adserver/models.py
```diff
@@ -291,33 +291,48 @@ def daily_reports(
...
-            ) + (impression.views * float(impression.advertisement.flight.cpm) / 1000.0)
-            days[index]["revenue_share"] = days[index]["revenue"] * (
-                self.revenue_share_percentage / 100.0
-            )
-            days[index]["our_revenue"] = (
-                days[index]["revenue"] - days[index]["revenue_share"]
-            )
+
+            if impression.advertisement:
+                days[index]["revenue"] += (
```
davidfischer, line 300: Should we set these to 0 in the `else` case?

davidfischer (approved):
I still think this is extremely awkward but we do need it. I would move the call to `Advertisement.incr` into `record_null_offer` but that's just me. Also, the migrations need to be reordered. Otherwise, I think we do want this despite its awkwardness.

adserver/models.py
ericholscher: We're doing a `+=`, so setting it to 0 will break things. I think this logic works -- at least it works fine in my testing. I guess it could break for a site with only Decisions and no Offers/Views

PR #279, opened 2020-11-09
Add uplift to Offers

ericholscher:
This allows us to track the uplift of a view.
It requires an updated ad client that appends `uplift` on view proxy images.


davidfischer (approved):
Seems simple enough. So we'll just add `?uplift=true` to the view pixel. Sounds good.

ericholscher:
> So we'll just add ?uplift=true to the view pixel. 

Yep, that's the idea anyway :)

PR #282, opened 2020-11-12
Fix offer refund logic

ericholscher:


davidfischer (commented):
Based on our discussion, I think we need a few more things here:

- The aggregation tasks need to disregard `offer.is_refunded` in their aggregations
- We need to re-run the aggregation tasks on the relevant dates

davidfischer:
> The aggregation tasks need to disregard offer.is_refunded in their aggregations

This part is done. I'd be OK merging this although when we do a refund we have to re-run the aggregation for the relevant dates.

davidfischer (approved):

PR #283, opened 2020-11-12
Tweak payout reporting to include only paid campaigns

ericholscher:
This gives us CTR from the paid campaigns, not all.


davidfischer (approved):
Seems simple enough.

PR #284, opened 2020-11-13
Deprecate ad types

davidfischer:
- Deprecated ad types can't be used with new ads (except by admins)
- Deprecated ad types can't be added to existing ads
- Users can continue to edit ads with deprecated ad types although once the deprecated ad type is removed, they can't re-add it
- Admin can filter by deprecated


adserver/views.py
```diff
@@ -286,7 +286,9 @@ def get_context_data(self, **kwargs):
             {
                 "advertiser": self.advertiser,
                 "flight": self.flight,
-                "ad_types": AdType.objects.exclude(description="")[:5],
+                "ad_types": AdType.objects.exclude(deprecated=True).exclude(
+                    description=""
+                )[:5],
```
ericholscher, line 291: Is there a reason to filter this by `:5`? Presumably we just want all of them?

ericholscher (approved):
Looks good -- having a deprecation path seems good 👍

adserver/views.py
davidfischer: Well, this is 5 with descriptions. If we had say 10 ad types with descriptions, there would be a lot of text on the right side. It isn't wrong but it probably isn't what we want.

PR #285, opened 2020-11-13
Additional payout automation

ericholscher:
This slowly automates more payout functions:

* Automatically send publisher a Front email
* Automatically create a Publisher Payout model

The next steps here would be to actually send the money to publishers via Stripe & Paypal API.
I'd say 90% of publishers use one of these options,
so we'll probably end up automating that more in the next couple months.


davidfischer (approved):
This looks fine although the linting issues (complexity, too long of a function) is mostly true. Perhaps a small refactor to have something like `handle_publisher_payout` and `send_publisher_email` or something like that.

PR #286, opened 2020-11-17
Add keyword reporting for publishers

ericholscher:
This builds a new KeywordImpression index,
which is used to show ads matching a specific keyword.
This is a somewhat naive approach,
but I think it's useful to know how any ad matches a keyword,
and I can't think of a way to not show 100%+ for those cases.


adserver/views.py
```diff
@@ -562,11 +562,6 @@ def get_context_data(self, **kwargs):
         start_date = self.get_start_date()
         end_date = self.get_end_date()
         campaign_type = self.request.GET.get("campaign_type", "")
-        revenue_share_percentage = self.request.GET.get("revenue_share_percentage", "")
-        div_id = self.request.GET.get("div_id", "")
-        # This needs to be something other than `advertiser` to not conflict with template context on advertising reports.
-        report_advertiser = self.request.GET.get("report_advertiser", "")
-        country = self.request.GET.get("country", "")
```
ericholscher, line 569: These were moved off the base onto the specific reports.

adserver/tasks.py
```diff
@@ -143,3 +148,48 @@ def daily_update_impressions(day=None):
...
+
+    for impression_type in IMPRESSION_TYPES:
+        queryset = _default_filters(impression_type, start_date, end_date)
+
+        for values in (
+            queryset.values("publisher", "advertisement", "keywords")
+            .annotate(total=Count("keywords"))
+            .filter(total__gt=0)
+            .order_by("-keywords")
+        ):
```
davidfischer, line 172: This gets the unique publisher/advertisement/keywords combination for a day. This is going to be enormously expensive especially for publishers like Read the Docs. Did you try this in production to see how long it takes and how many results it yields? I am worried about running this every 30 minutes.

adserver/tasks.py
```diff
@@ -143,3 +148,48 @@ def daily_update_impressions(day=None):
...
+
+            keywords = json.loads(values["keywords"])
+            ad = Advertisement.objects.filter(id=values["advertisement"]).first()
```
davidfischer, line 175: Since you use the flight later, you could save a query (in the loop) with a `select_related` here.

davidfischer (approved):
This looks fine. My big worry is the performance due to all the combinatorics.

adserver/tasks.py
ericholscher: It runs hourly, but we could also scale that back if we're worried about it. I haven't tested it against RTD.

ericholscher: I tested it with today's date in prod and it ran in ~10 seconds.

adserver/tasks.py
ericholscher: Thinking more on it, I think we might be able to return this data from the query. I'll see if that works reasonably well.

adserver/tasks.py
ericholscher: Removing the inline queries makes it really fast 👍

davidfischer:
Your improvements here look good. Removing the queries in the loop is great.

PR #287, opened 2020-11-17
Split Circle tests into two

ericholscher:
This makes it more obvious if tests or checks failed.
It also allows us to run them in parallel.


stsewd (approved):

davidfischer (approved):
We might be able to speed it up by caching the virtualenv but that's a double edged sword sometimes. LGTM.

PR #289, opened 2020-11-17
Fix keyword reporting by making it aggregate.

ericholscher:
This also includes a few other report fixes on fill and view rate,
and keyword storage.

### DIFF
--- a/adserver/models.py
+++ b/adserver/models.py
@@ -339,7 +339,7 @@ class Publisher(TimeStampedModel, IndestructibleModel):
             days[index]["fill_rate"] = calculate_ctr(
                 days[index]["offers"], days[index]["decisions"]
             )
-            days[index]["offer_rate"] = calculate_ctr(
+            days[index]["view_rate"] = calculate_ctr(
                 days[index]["views"], days[index]["offers"]
             )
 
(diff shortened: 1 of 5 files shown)


adserver/tasks.py
```diff
@@ -199,5 +206,5 @@ def daily_update_keywords(day=None):
                     keyword=keyword,
                 )
                 KeywordImpression.objects.filter(pk=impression.pk).update(
-                    **{impression_type: values["total"]}
+                    **{impression_type: F(impression_type) + values["total"]}
```
davidfischer, line 209: I understand the need for this but it's a bit confusing and different from elsewhere. It could probably use a comment. 

I'll just add this discussion for posterity:

> Let’s say you had 10 offers for the keywords “datascience, machine learning” and 5 offers for “datascience, jupyter” or something like that. You want to make sure you have 15 offers for “datascience”

adserver/models.py
```diff
@@ -350,7 +350,7 @@ def daily_reports(
                 days.values(), key=lambda obj: obj["views"], reverse=True
             )[:report_length]
 
-        report["total"]["decisions"] = sum(day["decision"] for day in report["days"])
+        report["total"]["decisions"] = sum(day["decisions"] for day in report["days"])
```
davidfischer, line 353: I'm working on making sure this is more easily testable.

davidfischer (approved):

PR #290, opened 2020-11-18
Don't overwrite Offer on uplift

ericholscher:
This was causing offer.viewed to be overwritten,
because we're using a specific Offer object instead of re-querying it.


davidfischer (approved):
Looks good but could probably use a test for this kind of critical path functionality.

ericholscher:
Yea -- I don't think we have great test coverage for this code. I will see about expanding it.

ericholscher:
We did have tests for this, but it didn't catch this specific issue. I've updated it to hopefully catch it in the future.

PR #294, opened 2020-11-24
Add uplift reporting

ericholscher:
This reuses the All Publishers report,
and just changes the index on the total.
This actually worked pretty cleanly.

Another option for this would be to add `uplift` to the AdImpression index.
The data is all the same,
and it's just an additional boolean that we'd pivot on.
We can always go that way if we want,
but this approach seemed simplest for now.


adserver/views.py
```diff
@@ -1414,17 +1417,6 @@ def get_context_data(self, **kwargs):  # pylint: disable=too-many-locals
         )
         publishers = Publisher.objects.filter(id__in=impressions.values("publisher"))
 
-        advertiser_list = (
```
ericholscher, line 1417: This wasn't being used.

adserver/tasks.py
```diff
@@ -211,3 +212,35 @@ def daily_update_keywords(day=None):
                 KeywordImpression.objects.filter(pk=impression.pk).update(
                     **{impression_type: F(impression_type) + values["total"]}
                 )
+
+
+@app.task()
+def daily_update_uplift(day=None):
```
davidfischer, line 218: Just a note that this task is not in celerybeat so it will not run automatically.

adserver/templates/adserver/reports/all-publishers_uplift.html
```diff
@@ -0,0 +1,15 @@
+{% extends "adserver/reports/all-publishers.html" %}
+{% load humanize %}
+{% load i18n %}
+
+
+{% block title %}{% trans 'Uplift Report' %}{% endblock %}
```
davidfischer, line 6: Should we add an explainer in here? I know this is just for us but it might be useful.

davidfischer (approved):
This worked well for me. I noted a few very minor things but this looks and works well.

adserver/tasks.py
ericholscher: Fixed.

PR #295, opened 2020-11-26
Advertiser reporting breakdowns

davidfischer:
Adds two reports for advertiser (although they're staff-only currently).

* An report that breaks down the top geos for an advertiser. Currently this doesn't allow filtering by flights so they can't see a flight by geos (yet). I'm not sure how needed that is.
* A report that breaks down the top publishers where an advertisers ads are shown. Again, this is across all ads/flights for the advertiser within the specified time period.

I still feel like there's some duplicated code in our reports and I hope to clean that up (possibly with django-filter or the like) in another PR.


adserver/views.py
```diff
@@ -751,6 +753,152 @@ def get_context_data(self, **kwargs):
         return context
 
 
+class AdvertiserGeoReportView(AdvertiserAccessMixin, GeoReportMixin, BaseReportView):
+
+    """A report for an advertiser broken down by geo."""
+
+    model = GeoImpression
+    template_name = "adserver/reports/advertiser-geo.html"
```
ericholscher, line 761: Should probably do `s/_/-/` for the publisher report paths, since I used underscores.

adserver/views.py
```diff
@@ -751,6 +753,152 @@ def get_context_data(self, **kwargs):
...
+            order="-date" if country else None,
+            max_results=None if country else self.LIMIT,
+        )
+        report.generate()
+
+        country_options = self.get_country_options(
+            self.get_queryset(
+                advertiser=advertiser,
+                start_date=context["start_date"],
+                end_date=context["end_date"],
```
ericholscher, line 794: This should probably use the `start_date` and `end_date` vars we already pulled out, or just access them directly in the context. I don't have a preference, just want it to be the same.

adserver/views.py
```diff
@@ -751,6 +753,152 @@ def get_context_data(self, **kwargs):
...
+        )
+
+        context.update(
+            {
+                "advertiser": advertiser,
+                "report": report,
+                "country": country,
+                "country_options": country_options,
+                "country_name": self.get_country_name(country),
+                "export_url": "{url}?{params}".format(
```
ericholscher, line 805: This feels a bit heavy to put inside the `context.update` -- should `get_export_url` live on the class?

adserver/views.py
```diff
@@ -751,6 +753,152 @@ def get_context_data(self, **kwargs):
...
+                    url=reverse("advertiser_geo_report_export", args=[advertiser.slug]),
+                    params=urllib.parse.urlencode(
+                        {
+                            "start_date": context["start_date"].date(),
+                            "end_date": context["end_date"].date()
+                            if context["end_date"]
+                            else "",
```
ericholscher, line 812: This if/else reads a bit weird to me here. Maybe just that it isn't indented?

adserver/views.py
```diff
@@ -751,6 +753,152 @@ def get_context_data(self, **kwargs):
...
+
+    """A report for an advertiser broken down by publishers where the advertisers ads are shown."""
+
+    template_name = "adserver/reports/advertiser-publisher.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        start_date = context["start_date"]
+        end_date = context["end_date"]
```
ericholscher, line 833: Similarly, I think we probably just remove these vars, and access it directly on the context.

adserver/views.py
```diff
@@ -751,6 +753,152 @@ def get_context_data(self, **kwargs):
...
+
+        advertiser_slug = kwargs.get("advertiser_slug", "")
+        advertiser = get_object_or_404(Advertiser, slug=advertiser_slug)
+        report_publisher = Publisher.objects.filter(
+            slug=self.request.GET.get("publisher", "")
```
ericholscher, line 838: We should probably change the GET arg on the publisher reports to just `advertiser`, instead of exposing `report_advertiser` in the URL heh.

ericholscher (approved):
Looks 💯 . I like the mixin with the common logic, that's definitely a nice improvement.

adserver/views.py
davidfischer: Yes, this looks a bit awkward to me as well although that's how black formatted it.

davidfischer:
This is ready for another look if needed.

ericholscher (approved):
Looks good! Yay for consistency 💯

PR #298, opened 2020-12-04
Fix a few issues with refunding

davidfischer:
- Refunding now *only* works on the Offer model
- Related clicks/views are also refunded

## Notes
- I'd consider moving `invalidate_nonce` to be called from `track_view` and `track_click`. I don't see any reason why those methods shouldn't call it. If agreed, I'll do that in a separate PR.
- Currently refunding doesn't update the relevant index tables (eg. GeoImpression). I noted that in the comments but perhaps another solution would be to make the admin action grab the unique dates and queue up re-indexing on those dates. Thoughts?
- Unfortunately, there isn't something tying an `Offer` object to a `View` or `Click` object. Those don't get "refunded" as part of this but I guess that's OK since we don't really use them for anything. We might want to clean that up at some point.

### DIFF
--- a/adserver/admin.py
+++ b/adserver/admin.py
@@ -714,7 +714,6 @@ class AdBaseAdmin(RemoveDeleteMixin, admin.ModelAdmin):
 
     """Django admin configuration for the base class of ad views and clicks."""
 
-    actions = ["refund_impressions"]
     readonly_fields = (
         "date",
         "advertisement",
@@ -726,7 +725,6 @@ class AdBaseAdmin(RemoveDeleteMixin, admin.ModelAdmin):
         "os_family",
         "is_mobile",
         "is_bot",
-        "is_refunded",
         "user_agent",
         "ip",
         "div_id",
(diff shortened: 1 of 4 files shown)


ericholscher (approved):
> I'd consider moving invalidate_nonce to be called from track_view and track_click. I don't see any reason why those methods shouldn't call it. If agreed, I'll do that in a separate PR.

Yea, this makes sense. 

> Currently refunding doesn't update the relevant index tables (eg. GeoImpression). I noted that in the comments but perhaps another solution would be to make the admin action grab the unique dates and queue up re-indexing on those dates. Thoughts?

Seems reasonable, though could be expensive if its over a bunch of days. Not sure the best approach -- having a bunch of invalid data isn't great either. 

> Unfortunately, there isn't something tying an Offer object to a View or Click object. Those don't get "refunded" as part of this but I guess that's OK since we don't really use them for anything. We might want to clean that up at some point.

Yea, I think mapping a Click to an Offer could be a useful change in the future.

PR #300, opened 2020-12-08
Add a backend option to exclude publishers for an advertiser

davidfischer:


adserver/models.py
```diff
@@ -305,6 +305,12 @@ class Campaign(TimeStampedModel, IndestructibleModel):
         ),
     )
 
+    exclude_publishers = models.ManyToManyField(
```
ericholscher, line 308: Is there a reason we have this on the Campaign and not at the Flight level? I guess we're doing the Campaign <-> Publisher mapping here as well with publisher groups, so it makes sense, but I definitely found it hard to follow the mapping of this stuff having it in a different place than all the other targeting information on the Flight.

ericholscher (approved):
Seems simple enough.

adserver/models.py
davidfischer: It doesn't matter too much but the way I think of it is that things that should almost never change go on the Campaign while things that might change from one ad buy to the next go on the flight. I'm open to moving it.

ericholscher: Not a huge deal, thanks for explaining the mental model, it makes sense. 👍

Issue #152, opened 2020-07-01
Break out ad headline from ad


davidfischer:
Currently, we allow some limited HTML formatting in our ads and I'd like to stop this and instead have a bolded headline followed by plain text. This would allow for easier validation and custom formatting by publishers. While we use a good HTML sanitizer, I think not allowing HTML at all is probably better for security.

This will involve:

- Migrating existing ads that use HTML to the new format
- Changing our database model to support a headline
- Possibly changing ad styles to support the new output

## Screenshot of existing

![Screen Shot 2020-07-01 at 11 57 56 AM](https://user-images.githubusercontent.com/185043/86281267-20e94800-bb92-11ea-8647-9f755c7b1a3c.png)

ericholscher:
I think this is a good idea, and should make image & text ads closer together in concept 👍

davidfischer:
Here's my plan here:

- All ads will have a body and headline. We should also maybe consider an optional CTA (thoughts?). The headline and CTA will be bolded in the output. The body, headline, and CTA will all be plain text with no HTML.
- For new ads being created, always use a form that breaks the fields out and doesn't use `ad.text`. Behind the scenes, `ad.text` will get constructed from the body, headline, and CTA. When editing existing ads, use the old form with just `ad.text`. On the model, `ad.body`, `ad.headline` will be null/empty in the latter case. The form will ensure the combined length of the body, headline, and CTA fit in the required space.
- The ad API response will have the fields broken out. We will attempt to calculate the headline, body, and CTA in a similar fashion to this: https://github.com/readthedocs/ethical-ad-server/blob/5cea748c12ed748a904ed3125adcd7c843031308/adserver/models.py#L944-L947
- I'm leaning toward not having any sort of data migration for old ads. They'll just slowly phase out.

## Example

For example, with the following ad:

```python
ad.headline = "My Company"
ad.body = "Our widgets are the best."
ad.cta = "Try free for 30 days!"
```

The resulting HTML for the ad would be:

```html
<strong>My Company</strong> Our widgets are the best. <strong>Try free for 30 days!</strong>
```
```

ericholscher:
Yea, this is what I was imagining. The API response contains the broken out fields, but also the `text` field which just standardizes the HTML response from the fields for legacy providers.

PR #302, opened 2020-12-14
Break the ad headline and CTA from the body

davidfischer:
This breaks the ad headline, body, and CTA out of the text.

* New ads will all require setting at least the body and optionally the headline and CTA. These new fields do not support HTML.
* Existing ads will continue to support the old interface with limited HTML supported. We can either migrate those old ads or just let them phase out.
* The only tricky part of rolling this out (I think) will be publishers who already use the `body` field in our API response. They will want to also take the `headline` and `cta` fields. For old ads, `headline` and `cta` will be `None`.

Fixes: https://github.com/readthedocs/ethical-ad-server/issues/152


ericholscher:
> The only tricky part of rolling this out (I think) will be publishers who already use the body field in our API response. They will want to also take the headline and cta fields. For old ads, headline and cta will be None.

This seems broken. We should just add a new field for body, and adjust the existing body copy to include the entire ad content without HTML. That is what it currently means, so we shouldn't break those integrations.

adserver/forms.py
```diff
@@ -156,10 +163,14 @@ def clean(self):
 
             # Check text length
             if ad_type.max_text_length:
-                stripped_text = bleach.clean(text, tags=[], strip=True)
+                if text:
+                    stripped_text = bleach.clean(text, tags=[], strip=True)
+                else:
+                    stripped_text = f"{headline} {body} {cta}"
```
ericholscher, line 169: This feels a bit weird. We should probably normalize the spaces to make sure it only includes 1 space.

adserver/models.py
```diff
@@ -941,15 +974,25 @@ def offer_ad(self, request, publisher, ad_type_slug, div_id, keywords):
...
+            body = html.unescape(bleach.clean(self.text, tags=[], strip=True))
+        if not text:
+            text = self.render_links(click_url)
 
         return {
             "id": self.slug,
-            "text": self.text,
+            "text": text,
+            "headline": headline,
             "body": body,
```
ericholscher, line 994: Maybe we migrate the new ad data into a dict?:

```
ad:
  - headline
  - body
  - cta
```
Then keep the existing top-level body as:

```
body: bleach.clean(headline + body + cta)
```

Or maybe we rename `body` in the new ad format, so we don't have confusing overloaded terms. Perhaps `headline, content, cta`?

ericholscher (commented):
This looks like a good change. I don't know if we want to block it on publishers adjusting their integration tho, especially given that it would be a weird chicken/egg problem on when they could deploy reading the new data. I think we should keep backwards compat on the API and just introduce the new data without breaking things.

adserver/forms.py
davidfischer: When the ad is actually rendered, this does happen correctly. This is only used to check that the ad length isn't too long.

adserver/models.py
davidfischer: I created a new dict named `copy` and put the 3 values in there.

davidfischer:
> This seems broken. We should just add a new field for body, and adjust the existing body copy to include the entire ad content without HTML. That is what it currently means, so we shouldn't break those integrations.

I ended up switching to what you suggested. It's a bit awkward still for folks just randomly taking `body` since it'll just be essentially "{headline} {content} {cta}" but I guess that's better than nothing.

ericholscher (approved):
These changes look good 👍

davidfischer:
I have to rebase this before merging to fix the migration versions.

PR #306, opened 2021-01-09
Add a null offer constraint

davidfischer:
This fixes an occasional race condition we see where two null offers for the same date/publisher. This can occur when two requests come in within a few milliseconds of each other immediately after midnight UTC. There are 23 offending records in our DB. This migration will probably lock the table although the migration shouldn't be too bad since this table only has ~300k rows in production.

### DIFF
--- a/adserver/models.py
+++ b/adserver/models.py
@@ -13,6 +13,7 @@ from django.core.validators import MinValueValidator
 from django.db import IntegrityError
 from django.db import models
 from django.db import transaction
+from django.db.models.constraints import UniqueConstraint
 from django.template import engines
 from django.template.loader import get_template
 from django.urls import reverse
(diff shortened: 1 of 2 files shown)


adserver/models.py
```diff
@@ -1227,6 +1228,14 @@ class AdImpression(BaseImpression):
     )
 
     class Meta:
+        # We must also constrain when the `advertisement` is null
+        constraints = (
+            UniqueConstraint(
```
ericholscher, line 1233: Cool, I didn't know Django could do this 👍

ericholscher (approved):
Will be good to get this fixed, thanks for actually doing it, instead of just researching it and then not doing anything, like me :)

adserver/models.py
davidfischer: It's new in 2.2.

PR #309, opened 2021-01-21
Small admin improvements

davidfischer:
- The big one here is searching advertisers
- Some simple records should just display more per page


ericholscher (approved):
Heh, I have the full publisher list bookmarked :)

This is better 👍

PR #311, opened 2021-01-22
Add a management command for creating advertisers

davidfischer:


adserver/management/commands/add_advertiser.py
```diff
@@ -0,0 +1,206 @@
...
+
+User = get_user_model()
+
+
+class Command(BaseCommand):
+
+    """Add an advertiser from the command line."""
+
+    DEFAULT_CPM = 3.33
+    DEFAULT_NUM_IMPRESSIONS = 150000
```
ericholscher, line 46: I wonder if we should make these configurable in the creation step (eg. `--plan starter` or `--cpm 3.5`). Probably a v2 feature, since it will probably require customization regardless.

adserver/management/commands/add_advertiser.py
```diff
@@ -0,0 +1,206 @@
...
+        "NZ",
+    ]
+
+    def add_arguments(self, parser):
+        # These fields will prompt if not provided
+        parser.add_argument(
+            "-e", "--email", type=str, help=_("Email of user managing this advertiser")
+        )
+        parser.add_argument(
+            "-n", "--name", type=str, help=_("Name of user managing this advertiser")
```
ericholscher, line 81: I haven't been adding these for publishers. I guess its useful if we ever want to automatically send them email from the app?

adserver/management/commands/add_advertiser.py
```diff
@@ -0,0 +1,206 @@
...
+        )
+        parser.add_argument(
+            "--num-impressions",
+            type=int,
+            default=self.DEFAULT_NUM_IMPRESSIONS,
+            help=_(
+                "Number of impressions for the first flight [%s]"
+                % self.DEFAULT_NUM_IMPRESSIONS
+            ),
+        )
```
ericholscher, line 102: Ah, you already did it :)

ericholscher (approved):
Looks good 👍

adserver/management/commands/add_advertiser.py
davidfischer: It's slightly more important here because we use the same name on Stripe.

PR #312, opened 2021-02-02
Show whats left on a flight always

davidfischer:


ericholscher (approved):
💯  -- small, solid improvement

PR #314, opened 2021-02-03
Don't count ad display when a particular ad is forced

davidfischer:


adserver/models.py
```diff
@@ -971,7 +973,12 @@ def offer_ad(self, request, publisher, ad_type_slug, div_id, keywords):
             ad_type_slug=ad_type_slug,
         )
 
-        nonce = offer.pk
+        if forced:
+            # Ad offers forced to a specific ad or campaign should never be billed
+            # By discarding the nonce, the ad view/click will never count
+            nonce = "forced"
```
ericholscher, line 979: Clever way to handle this 👍

ericholscher (approved):

PR #318, opened 2021-02-22
Run previous days reports automatically

davidfischer:
Add a scheduled task to run all of the reports for the previous day.
This needs to be run after the day is complete or it might miss some data.

### DIFF
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -251,7 +251,7 @@ def daily_update_uplift(day=None):
 
 
 @app.task()
-def daily_update_all_reports(day=None):
+def update_previous_day_reports(day=None):
     """
     Complete all report data for the previous day.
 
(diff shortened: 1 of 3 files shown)


adserver/tasks.py
```diff
@@ -248,3 +248,23 @@ def daily_update_uplift(day=None):
             UpliftImpression.objects.filter(pk=impression.pk).update(
                 **{impression_type: values["total"]}
             )
+
+
+@app.task()
+def daily_update_all_reports(day=None):
```
ericholscher, line 254: A bit confusing to pass in a day and index the prior. Probably need to change this name a bit:

```suggestion
def update_previous_day_reports(day=None):
```

ericholscher (approved):
Good idea. 👍

adserver/tasks.py
davidfischer: I'll take this change although I have to change more than this one line.

PR #319, opened 2021-02-24
Uplift report updates

davidfischer:
- Make publisher breakdown 1 line per publisher
- Allow exporting

## Screenshot
Screenshot from 2021-02-24 16-29-05

### DIFF
--- a/adserver/views.py
+++ b/adserver/views.py
@@ -616,6 +616,8 @@ class BaseReportView(UserPassesTestMixin, TemplateView):
         # Publisher filters
         if "publisher" in kwargs and kwargs["publisher"]:
             queryset = queryset.filter(publisher=kwargs["publisher"])
+        if "publishers" in kwargs and kwargs["publishers"]:
+            queryset = queryset.filter(publisher__in=kwargs["publishers"])
         if "campaign_type" in kwargs and kwargs["campaign_type"] in ALL_CAMPAIGN_TYPES:
             queryset = queryset.filter(
                 advertisement__flight__campaign__campaign_type=kwargs["campaign_type"]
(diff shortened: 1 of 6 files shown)


ericholscher:
Yep, this looks great!

davidfischer:
> Yep, this looks great! If it isn't a huge pain, the total should probably go at the bottom of the publisher breakdown, so there's only 1 list, and the total is at the bottom?

You're thinking then to remove the breakdown by day? Right now, the total is at the bottom of the daily breakdown.

ericholscher:
Gotcha, that sounds 💯

davidfischer:
I updated the report (and updated the screenshot in the PR). The report now shows a breakdown by publisher only, no daily breakdown. There's also an export button.

adserver/views.py
```diff
@@ -1536,15 +1538,72 @@ def get_context_data(self, **kwargs):  # pylint: disable=too-many-locals
...
+        revenue_share_percentage = self.request.GET.get("revenue_share_percentage", "")
+        if revenue_share_percentage:
+            try:
+                publishers = publishers.filter(
+                    revenue_share_percentage=float(revenue_share_percentage)
+                )
+            except ValueError:
+                pass
+
+        queryset = self.get_queryset(
```
ericholscher, line 1574: Calling `get_queryset` 2x in this function is a bit weird, I'm guessing just because we're filtering for publishers then doing the full data set?

ericholscher (approved):
Looks good 👍

adserver/views.py
davidfischer: Ya. We could probably get rid of the 2nd call.

davidfischer: I removed it

PR #320, opened 2021-02-25
Fix off by 1 (actually 2) error in ad text size

davidfischer:
I got a report of this from an advertiser. Basically if they leave the headline or CTA blank it can cause the total text length to be off slightly.


ericholscher (approved):
Heh, interesting. I did see some times in ad display where the content & CTA don't have a period between them, which looks a bit weird. Not sure if we want to normalize this across everything, but definitely somewhere that things look a bit odd.

PR #324, opened 2021-03-05
Add debug flag to payout command

ericholscher:
This makes it less verbose unless we want the debug info

### DIFF
--- a/adserver/management/commands/payouts.py
+++ b/adserver/management/commands/payouts.py
@@ -116,6 +116,13 @@ class Command(BaseCommand):
             required=False,
             action="store_true",
         )
+        parser.add_argument(
+            "-d",
+            "--debug",
+            help="Print debug output",
+            required=False,
+            action="store_true",
+        )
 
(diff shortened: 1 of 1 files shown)


adserver/management/commands/payouts.py
```diff
@@ -213,7 +225,7 @@ def handle(self, *args, **kwargs):
                     "to": [user.email for user in publisher.user_set.all()],
                     "sender_name": "EthicalAds by Read the Docs",
                     "subject": f"EthicalAds Payout - {publisher.name}",
-                    "options": {"archive": False},
+                    "options": {"archive": True},
```
davidfischer, line 228: Is this an inadvertent change?

davidfischer (approved):

adserver/management/commands/payouts.py
ericholscher: Nope -- it was spamming my inbox, so I'm archiving the messages by default.

PR #325, opened 2021-03-05
Calculate Fill Rate against only paid offers

ericholscher:
This should fix the display of fill rate


davidfischer (approved):
That ended up being not as bad as I feared.

PR #327, opened 2021-03-05
Add basic stats to the dashboard view

ericholscher:
This adds the total publisher data to the view.
I made a basic UI for it,
but probably needs to be cleaned up.

This currently breaks tests, but I wanted to know if we wanted to put these stats at the homepage, or always redirect users to their publisher account in they only have 1, and put them there? Seems best to me to have them always on the homepage, to simplify logic.

### DIFF
--- a/adserver/templates/adserver/dashboard.html
+++ b/adserver/templates/adserver/dashboard.html
@@ -16,6 +16,29 @@
 
   {% if publishers %}
     <h2>{% trans 'Publishers' %}</h2>
+    <div class="container">
+      <div class="row">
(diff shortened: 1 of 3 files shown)


davidfischer:
> This currently breaks tests, but I wanted to know if we wanted to put these stats at the homepage, or always redirect users to their publisher account in they only have 1, and put them there? Seems best to me to have them always on the homepage, to simplify logic.

I was going to propose the opposite. I think having an overview on the publisher/advertiser specific page is a bit better rather than trying to aggregate it or something.

PR #323, opened 2021-03-05
Publisher report cleanup

ericholscher:
### Changes

* stops showing publishers our revenue
* moves all the "staff" reports to the right and puts a visual separator.
* Tweaks the month display in the reports to always use 3 letter months
* Use a monospace font

### Questions

* Not 100% sure on the best font to use. Could use some feedback there :)
* Should we show the eCPM to publishers?
* Should we show fill rate to publishers? We had a request for this
  - Fill rate is currently not very useful if they have house ads turned on, so probably need to fix that first

### Preview

Screen Shot 2021-03-04 at 5 57 24 PM

PR #328, opened 2021-03-05
Total revenue report improvements

davidfischer:
These are some suggestions for #323.

- Use CSS class report for reports
- Use monospace font only for reports
- Display header for staff-only columns in a different color


ericholscher (approved):
Looks great! 👍

PR #329, opened 2021-03-08
Remove refunded offers from aggregate reports

davidfischer:
Refunds are extremely rare but cause report generation to be *much* slower. We will tolerate slightly inaccurate reports (the actual $ spent reports will be correct) to have them be created faster.

### DIFF
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -46,7 +46,7 @@ def _default_filters(impression_type, start_date, end_date):
     queryset = Offer.objects.filter(
         date__gte=start_date,
         date__lt=end_date,  # Things at UTC midnight should count towards tomorrow
-        is_refunded=False,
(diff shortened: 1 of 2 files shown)


adserver/tasks.py
```diff
@@ -46,7 +46,7 @@ def _default_filters(impression_type, start_date, end_date):
     queryset = Offer.objects.filter(
         date__gte=start_date,
         date__lt=end_date,  # Things at UTC midnight should count towards tomorrow
-        is_refunded=False,
+        # is_refunded=False,  # This causes the query to be a filtered index and is much slower
```
ericholscher, line 49: We can always make this a kwarg if we want, when we actually have to offer refunds.

ericholscher (approved):

adserver/tasks.py
davidfischer: If this ends up being a good performance fix, we'll do that.

PR #332, opened 2021-03-12
Archive offers management command

davidfischer:
- Archives offers to CSVs
- Can write those CSVs to backups
- Can run a delete afterwards (requires an option)

### DIFF
--- a/adserver/management/commands/archive_offers.py
+++ b/adserver/management/commands/archive_offers.py
@@ -181,11 +181,15 @@ class Command(BaseCommand):
 
         # Because the last day is inclusive, the delete query needs to delete after that day
         end_day += datetime.timedelta(days=1)
+        query = "DELETE FROM adserver_offer WHERE date >= %s AND date < %s"
+
+        self.stdout.write(_("- Executing SQL:"))
+        self.stdout.write(query % (start_day, end_day))
 
         with connection.cursor() as cursor:
             # Offers are normally immutable so the delete has to be run as raw sql
             cursor.execute(
-                "DELETE FROM adserver_offer WHERE date >= %s AND date < %s",
+                query,
                 [start_day, end_day],
             )
             deleted_offers = cursor.fetchone()
@@ -193,6 +197,9 @@ class Command(BaseCommand):
         self.stdout.write(
             self.style.SUCCESS(_("Successfully removed %d offers.") % deleted_offers)
         )
+
+        # Update the DB stats if we removed offers
+        self.update_db_stats()
 
     def update_db_stats(self):
         """Updates DB stats after these changes."""
@@ -226,4 +233,3 @@ class Command(BaseCommand):
         self.copy_offer_dumps()
         if kwargs["delete_offers"]:
             self.delete_offers(kwargs["start_date"], kwargs["end_date"])
-        self.update_db_stats()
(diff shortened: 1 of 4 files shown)


adserver/management/commands/archive_offers.py
```diff
@@ -0,0 +1,229 @@
...
+
+class Command(BaseCommand):
+
+    """Management command to help archive offers."""
+
+    help = "Archives offers in CSV form from the database to files: one per day."
+
+    default_output_dir = "/tmp"
+    default_end_date = (timezone.now() - datetime.timedelta(days=90)).date()
+    default_start_date = default_end_date - datetime.timedelta(days=30)
```
ericholscher, line 33: Are we thinking of running this nightly now? Could be good to keep the DB performance consistent.

adserver/management/commands/archive_offers.py
```diff
@@ -0,0 +1,229 @@
...
+        """Deletes offers from the database (requires them to be copied to settings.BACKUPS_STORAGE)."""
+        if not hasattr(settings, "BACKUPS_STORAGE"):
+            self.stdout.write(
+                self.style.WARNING(
+                    _("Skipping deleting archived offers (backups weren't copied)...")
+                )
+            )
+            return
+
+        self.stdout.write(_("Deleting archived offers..."))
```
ericholscher, line 180: We should log the dates here to be confident it's doing the right thing. We should probably also do one additional confirmation `if input() == 'y'`, since deleting is scary.

adserver/management/commands/archive_offers.py
```diff
@@ -0,0 +1,229 @@
...
+                SELECT * FROM adserver_offer
+                WHERE date >= '{day:%Y-%m-%d}' AND date < '{end_day:%Y-%m-%d}'
+                ORDER BY date
+            ) TO STDOUT WITH CSV HEADER"""
+        with connection.cursor() as cursor:
+            with open(output_file, "w") as fd:
+                # https://www.psycopg.org/docs/cursor.html#cursor.copy_expert
+                cursor.copy_expert(query, fd)
+
+        # This will be off by one because the CSV contains a header row
```
ericholscher, line 101: Off by 1 compared to what? The number output by the SQL command?

adserver/management/commands/archive_offers.py
```diff
@@ -0,0 +1,229 @@
...
+            default=False,
+            help=_(
+                "Delete offers after archiving (requires successful copying to remote backups)"
+            ),
+        )
+        parser.add_argument(
+            "-s",
+            "--start-date",
+            default=self.default_start_date,
+            type=self._from_isoformat,
```
ericholscher, line 70: Didn't know about that, neat.

ericholscher (approved):
Looks good 👍

adserver/management/commands/archive_offers.py
davidfischer: It requires a command line option `-d` to delete offers. The default is to *not* do it. Adding an input here means it can never be run automatically and it requires input after potentially many minutes (hours?) of doing the actual archiving. Logging the dates, however, seems reasonable.

adserver/management/commands/archive_offers.py
davidfischer: `wc -l` is outputting the lines in the file which will be one more than the actual number of records. The SQL copy doesn't output a number.

adserver/management/commands/archive_offers.py
davidfischer: I was still thinking maybe weekly or monthly. It shouldn't take nearly as much IO to do it day by day.

PR #338, opened 2021-03-19
Enable sticky ad decisions

davidfischer:
We want to make ad decisions a bit sticky. This means that if a user requests an ad for the same publisher and ad type in short succession, they will get the same ad. This helps with attribution of ads and doesn't result in wasted ad views.


adserver/api/views.py
```diff
@@ -148,43 +151,52 @@ def _prepare_response(self, ad, placement, publisher, keywords, forced=False):
...
-        # Record a decision for every call to the API
-
         ad_type_slug = placement.get("ad_type")
         div_id = placement.get("div_id")
 
-        if not ad:
-            Advertisement.record_null_offer(
+        # Check if this client should get a sticky ad decision
+        cache_key = self._sticky_decision_cache_key(publisher, ad_type_slug)
+        data = cache.get(cache_key)
```
ericholscher, line 159: I think we should probably log here if we're using a cached response, so we have a bit more data around this. INFO level probably to start, then we can drop it to DEBUG?

ericholscher (approved):
This seems like a solid improvement :+1:

adserver/api/views.py
davidfischer: Added

PR #346, opened 2021-03-22
Changes needed now that cryptography requires rust

davidfischer:
This only really affects development. In production (Ubuntu), apparently new versions of `pip` will "just work"


ericholscher (approved):

ericholscher:
I think it probably makes sense to migrate this to Ubuntu, to match prod at some point.

PR #348, opened 2021-03-25
Add publisher query parameter to ad clicks

davidfischer:
This appends a publisher query parameter `ea-publisher=<publisher-slug>` to ad click redirects.

## Proposed alternative

This does feel a bit like "magic" to me since the advertiser doesn't control it. Could this extra query parameter muck up the analytics that our advertisers are running? Would they prefer it in another variable? Another alternative would be to just document our existing substitution like so:

Screenshot from 2021-03-25 15-49-40


ericholscher (approved):
I think this is good. 

We should mention this is happening in the help text, instead of asking publishers to use some wonky syntax to add it themselves. That seems way more error prone.

Alternatively, we could have an additional field that is `query param for publisher data` and let advertisers customize this, default to `ea-publisher`, if we're worried they want a different parameter for it. I think this will be fine for now though.

davidfischer:
I added some help text but otherwise this is good to go.

ericholscher:
@davidfischer did you want to only do it when the user hadn't defined it, or just worry about that later?

davidfischer:
I'm on the fence. Just to be clear, you're thinking to only add `ea-publisher` if the user doesn't have `${publisher}` in the URL? I think I'm on board with that but it just feels like a bit of magic.

ericholscher:
Yea -- might be better to just always include it. Easier to understand. I can imagine a case where a subset of flights had a `publisher` and they got weird/missing data.

PR #351, opened 2021-04-01
Reporting UX improvements

davidfischer:
- Make each filter a bit wider and more responsive on smaller screens
- Store start and end date for reporting in the session (GET vars reset the session value). This means that when somebody clicks a link to the geo report or another report, it remembers the start and end date from the previous report.


ericholscher (approved):
That seems nice and clean 👍

PR #355, opened 2021-04-05
Tweaks to the archive management command

davidfischer:
The big change here is to copy a day's worth of offers to backups before moving on to the next day. Previously, it did all the archiving before doing any copying to backups. This makes the command easier to stop and start as needed

### DIFF
--- a/adserver/management/commands/archive_offers.py
+++ b/adserver/management/commands/archive_offers.py
@@ -131,7 +131,9 @@ class Command(BaseCommand):
 
         self.stdout.write(self.style.SUCCESS(_("Successfully archived %s.") % day))
 
-    def copy_offer_dumps(self):
+        return zipped_output_file
+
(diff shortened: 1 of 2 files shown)


adserver/management/commands/archive_offers.py
```diff
@@ -177,30 +172,26 @@ def delete_offers(self, start_day, end_day):
             )
             return
 
-        self.stdout.write(_("Deleting archived offers..."))
+        self.stdout.write(_("Deleting archived offers for %s...") % day)
```
ericholscher, line 175: This makes me less worried, since it's more bounded 👍

ericholscher (approved):
👍

adserver/management/commands/archive_offers.py
davidfischer: I probably still won't run it with that option until this whole process is very smooth.

PR #356, opened 2021-04-05
Create security policy

davidfischer:


ericholscher (approved):
Looks good. I wonder if we want a generic publicly accessible form for support inbound for stuff like this. We could exclude it from search engines, but I do dislike depending on email for this since it's very flakey.

davidfischer:
Ya, email is flakey. We could have folks open an issue on the GitHub repository. [HackerOne](https://www.hackerone.com/) has a product to handle this kind of thing although I seem to remember it was kind of expensive. Pricing is no longer public.

PR #358, opened 2021-04-07
Show CTR in summaries for ads and flights

davidfischer:
- This would make comparing ad performance easier
- This also makes comparing different targeting easier

Screenshot from 2021-04-06 21-28-32


ericholscher (approved):
Seems like a good change. I'd love to get View Time on here as well :)

davidfischer:
> I'd love to get View Time on here as well :)

Me too once it's done. I guess we should start scoping that. I'll create an issue.

PR #360, opened 2021-04-07
Add naive attempt at price targeting

ericholscher:
This adds a priority boost for ads that cost more money and have good CTR.
It should show more ads that make us more money over time,
without totally penalizing poor performing ads.

### DIFF
--- a/adserver/models.py
+++ b/adserver/models.py
@@ -650,9 +650,10 @@ class Flight(TimeStampedModel, IndestructibleModel):
 
         # Keep values above 1, so we don't penalize the value for lower performance
         # but add value for higher performance
-        value = max(int(ctr_value), 1)
+        value = max(float(ctr_value), 1.0)
+        value = min(value, 10.0)
 
-        return impressions_needed * self.priority_multiplier * value
+        return int(impressions_needed * self.priority_multiplier * value)
 
     def clicks_remaining(self):
         return max(0, self.sold_clicks - self.total_clicks)
--- a/adserver/models.py
+++ b/adserver/models.py
@@ -648,8 +648,8 @@ class Flight(TimeStampedModel, IndestructibleModel):
         # Leading to a base value of $2 * 5 * .1 == 1.0
         ctr_value = cost * 5.0 * float(self.ctr())
 
-        # Keep values above 1, so we don't penalize the value for lower performance
-        # but add value for higher performance
+        # Keep values between 1-10 so we don't penalize the value for lower performance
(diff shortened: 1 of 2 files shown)


adserver/models.py
```diff
@@ -632,14 +632,27 @@ def weighted_clicks_needed_today(self):
...
         impressions_needed += math.ceil(self.views_needed_today() / 1000.0)
         impressions_needed += self.clicks_needed_today()
 
-        return impressions_needed * self.priority_multiplier
+        # Cost should be a float from around .5-5, averaging 2
+        cost = float(self.cpm or self.cpc)
+
+        # CTR should be a float around .1
+        # Leading to a base value of $2 * 5 * .1 == 1.0
+        ctr_value = cost * 5.0 * float(self.ctr())
```
davidfischer, line 649: So if I'm reading this right, a $4 ad with a CTR of 0.2% would be ~4x as likely to be shown. That seems reasonable.

adserver/models.py
```diff
@@ -632,14 +632,27 @@ def weighted_clicks_needed_today(self):
...
+
+        # Keep values above 1, so we don't penalize the value for lower performance
+        # but add value for higher performance
+        value = max(int(ctr_value), 1)
```
davidfischer, line 653: I wouldn't mind having a maximum that to this multiplier. Say 10x or something like that. There are weird extremes that can happen like say the first time an ad is shown it's clicked resulting in a CTR of 100%. It's likely to even out pretty quickly but might be worth just capping so we don't run into that.

Also, we should probably keep the value priority multiplier as a float and only cast the final return value to an int.

```suggestion
        value = max(float(ctr_value), 1.0)
        value = min(value, 10.0)
```

adserver/models.py
```diff
@@ -632,14 +632,27 @@ def weighted_clicks_needed_today(self):
...
+
+        return impressions_needed * self.priority_multiplier * value
```
davidfischer, line 655:
```suggestion
        return int(impressions_needed * self.priority_multiplier * value)
```

davidfischer (approved):
I think the simplicity here is pretty good actually. It's a naive solution but probably being a little naive is where we want to be at first. We can always build something a bit more sophisticated over time.

ericholscher:
@davidfischer I added a few more tests so we have a bit more coverage here (cost, ctr, and bounds checking). I think I'm 👍 on it if you are.

davidfischer (approved):
Ship it!

PR #362, opened 2021-04-14
Mute the disallowed host logger in prod

davidfischer:
Similar to https://github.com/readthedocs/readthedocs.org/pull/4879.


ericholscher (approved):

davidfischer:
This doesn't seem to have worked correctly as I'm still seeing these...

Issue #354, opened 2021-04-02
Send URL with the ad request


davidfischer:
The URL where the ad is going to be displayed is useful for a few different reasons:

- We are considering building some functionality to process URLs after the fact to learn about what the content of those pages is about so we can have better contextual targeting
- Fraud detection
- Previously, we relied on the referrer which can't be trusted and quite a few of our publishers have a referrer policy that doesn't send it.

The main gotcha with collecting the URL is that we don't want to send any sensitive information in the URL like tokens. Maybe we just won't send query parameters? That won't be perfect on some sites but might be OK most of the time. Maybe we could have a blocklist of specific query parameters but that isn't ideal.

ericholscher:
I'm 👍  on this. It would also be useful for helping understand a publishers traffic and where it's coming from. Currently we're depending on placements, but if we have more URL-specific data it could help on the backend for a few other reporting views at some point. 

I think a blocklist of URL params is probably a good start (eg. `session`). We might still end up with a username or similar in the URL, but that's already an issue with referral data as well. 

> The Referer header may not contain URL fragments (i.e. "#section") or "username:password" information. It can potentially contain an origin, path, and querystring. What is sent, if anything, depends on the referrer policy for the request. See Referrer-Policy for information and examples.

I think we can probably use [similar logic](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer), so it works in the same fashion.

PR #365, opened 2021-04-17
Add an optional URL to the decision API

davidfischer:
This will allow the ad client to send the URL where the ad will appear. While the client can't be trusted, this will give more granular data for well-behaved clients.

Fixes https://github.com/readthedocs/ethical-ad-server/issues/354 (from the server side - client still needs to do this)

### DIFF
--- a/adserver/tests/test_api.py
+++ b/adserver/tests/test_api.py
@@ -1234,35 +1234,35 @@ class AdvertisingIntegrationTests(BaseApiTest):
         )
 
     def test_offer_url(self):
-        url1 = "http://example.com/path.html"
-        url2 = "http://example.com/altpath.html"
+        referrer_url = "http://example.com/path.html"
+        post_url = "http://example.com/altpath.html"
 
         # Defaults to the referrer if no URL sent
         resp = self.client.post(
             self.url,
             json.dumps(self.data),
             content_type="application/json",
-            HTTP_REFERER=url1,
+            HTTP_REFERER=referrer_url,
         )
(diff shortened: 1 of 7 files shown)


adserver/tests/test_api.py
```diff
@@ -1233,6 +1233,37 @@ def test_nullable_offers(self):
             1,
         )
 
+    def test_offer_url(self):
+        url1 = "http://example.com/path.html"
+        url2 = "http://example.com/altpath.html"
```
ericholscher, line 1238: Could be clearer to name this `referer_url` and `post_url` or similar.

adserver/tests/test_models.py
```diff
@@ -368,6 +368,39 @@ def test_projected_total_value(self):
...
+        request = self.factory.get("/")
+        request.ip_address = "1.1.1.1"
+        request.user_agent = (
+            "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) "
+            "Chrome/90.0.4430.72 Safari/537.36"
+        )
+
+        div_id = "foo"
+        url = "http://example.com/path.html"
+        keywords = ["python", "ruby"]
```
ericholscher, line 381: Should we confirm these are saved somewhere?

ericholscher (approved):
Looks like a solid change 👍

davidfischer:
I updated this with feedback.

ericholscher (approved):
Looks great 👍

PR #368, opened 2021-04-21
Don't reject invalid values in the URL field

davidfischer:
Make the URL field not reject invalid values. Instead, it will validate it and disregard an invalid value. The ad will still get offered. It does still reject a blank value (`&url=`).

After doing this, I'm not 100% sure we even want this. For other fields (eg. publisher) we actually do reject invalid values.

### DIFF
--- a/adserver/api/serializers.py
+++ b/adserver/api/serializers.py
@@ -1,4 +1,6 @@
 """De-serializers for the ad server APIs."""
+from django.core.exceptions import ValidationError
+from django.core.validators import URLValidator
 from rest_framework import serializers
 
 from ..constants import ALL_CAMPAIGN_TYPES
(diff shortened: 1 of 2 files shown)


adserver/api/serializers.py
```diff
@@ -82,6 +85,14 @@ def validate_keywords(self, keywords):
 
         return keywords
 
+    def validate_url(self, url):
+        validator = URLValidator()
+        try:
+            validator(url)  # Throws ValidationError on invalid
+            return url
+        except ValidationError:
+            return None
```
ericholscher, line 94: I think lets keep this logic, but log here to understand how often this is happening? Once we understand if its common, we can figure out what to do.

ericholscher (approved):
I don't think we want to reject invalid URL's quite yet without knowing what might cause them. Maybe lets report it to Sentry for now, and see how often we get it? It might just not really be an issue in practice.

davidfischer:
Sounds good. I added logging but it doesn't log on a blank URL (rejected) or not passed (simply OK).

PR #370, opened 2021-04-23
Show upcoming flights on the overview screen

davidfischer:
I noticed a few issues with the overview screen and this attempts to correct them. Firstly, it shows a client's upcoming ad flights. This is important especially for a new client. Otherwise they get a screen that doesn't show much and doesn't show anything upcoming.

Secondly, and this is really minor, it sets the `end_date` when clicking through from the overview screen to the last day of the month. If the user has a cookied `end_date` and clicks through, they can get into a slightly odd state where it doesn't show the whole month to date.

### DIFF
--- a/adserver/views.py
+++ b/adserver/views.py
@@ -1,4 +1,5 @@
 """Ad server views."""
+import calendar
 import collections
 import csv
 import logging
@@ -14,7 +15,7 @@ from django.contrib.auth.decorators import login_required
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.contrib.auth.mixins import UserPassesTestMixin
 from django.core.exceptions import ValidationError
-from django.db.models import Sum
+from django.db import models
 from django.http import Http404
 from django.http import HttpResponse
 from django.http import HttpResponseRedirect
@@ -40,6 +41,8 @@ from user_agents import parse as parse_user_agent
 
 from .constants import CAMPAIGN_TYPES
 from .constants import CLICKS
+from .constants import FLIGHT_STATE_CURRENT
+from .constants import FLIGHT_STATE_UPCOMING
 from .constants import VIEWS
 from .forms import AdvertisementForm
 from .forms import PublisherSettingsForm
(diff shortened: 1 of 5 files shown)


adserver/views.py
```diff
@@ -159,8 +162,11 @@ class AdvertiserMainView(
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
-        # Get the beginning of the month so we can show month-to-date stats
+        # Get the beginning/end of the month so we can show month-to-date stats
         start_date = timezone.now().replace(day=1, hour=0, minute=0, second=0)
+        end_date = start_date.replace(
+            day=calendar.monthrange(start_date.year, start_date.month)[1]
```
ericholscher, line 168: Huh, didn't know about this. Time is hard. I wish there was a more explicit way to get the data here other than [1], but alas.

adserver/views.py
```diff
@@ -169,18 +175,23 @@ def get_context_data(self, **kwargs):
...
-            .order_by("-live", "-end_date", "name")
-        )
+        flights = [
+            f
+            for f in (
+                Flight.objects.filter(campaign__advertiser=self.advertiser).order_by(
+                    "-live", "-end_date", "name"
+                )
+            )
+            if f.state in (FLIGHT_STATE_UPCOMING, FLIGHT_STATE_CURRENT)
```
ericholscher, line 185: Hrm. Can we not do this with the queryset? `state__in= [FLIGHT_STATE_UPCOMING, FLIGHT_STATE_CURRENT]`?

ericholscher (approved):
Seems useful 👍

adserver/views.py
davidfischer: Ya, I guess that's probably better.

davidfischer: Oh, no we can't. It's a calculated property.

ericholscher: gotcha, this works then 👍

adserver/views.py
davidfischer: Ya, I don't love it either.

PR #371, opened 2021-04-27
Copy/Re-use an existing ad

davidfischer:
This allows easy copying of an existing ad. I'm not completely sure on this UX but otherwise this should work well.

## Screenshot
Screenshot from 2021-04-27 16-35-10

### DIFF
--- a/adserver/views.py
+++ b/adserver/views.py
@@ -432,8 +432,7 @@ class AdvertisementCopyView(AdvertiserAccessMixin, UserPassesTestMixin, Template
         )
 
     def copy_instance(self):
-        instance = self.source_advertisement
-        instance.__copy__()
+        instance = self.source_advertisement.__copy__()
         instance.flight = self.flight
         instance.save()  # Automatically gets a new slug
         return instance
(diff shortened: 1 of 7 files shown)


adserver/views.py
```diff
@@ -364,6 +364,94 @@ def get_success_url(self):
...
+                "flight_detail",
+                kwargs={
+                    "advertiser_slug": self.advertiser.slug,
+                    "flight_slug": self.flight.slug,
+                },
+            )
+        )
+
+    def copy_instance(self):
+        instance = self.source_advertisement.__copy__()
```
ericholscher, line 435: Do we want to overwrite a python builtin here? Should we just call it `copy()`? Is not, should we use the python `copy` module to actually copy it?

ericholscher (approved):
Looks good. I didn't actually run it, so can't really comment on the UX. Good enough for now, based on the screenshot it seems.

adserver/views.py
davidfischer: I'm generally in favor of defining builtins when we are overriding that functionality. I don't have a super strong opinion on it.

PR #373, opened 2021-04-29
Allow advertisers to control their authorized users

davidfischer:
This gives advertisers the ability to control the authorized users who can access their reports and control ads. A few advertisers have multiple authorized users and have requested the ability to add more.

## Screenshot
Screenshot from 2021-04-29 09-41-30

### DIFF
--- a/adserver/forms.py
+++ b/adserver/forms.py
@@ -13,6 +13,7 @@ from crispy_forms.layout import Layout
 from crispy_forms.layout import Submit
 from django import forms
 from django.conf import settings
+from django.contrib.auth import get_user_model
 from django.core.files.images import get_image_dimensions
 from django.db.models import Q
 from django.urls import reverse
@@ -29,6 +30,7 @@ from .models import Publisher
 
 
 log = logging.getLogger(__name__)  # noqa
+User = get_user_model()
 
 
 class FlightAdminForm(forms.ModelForm):
@@ -391,4 +393,40 @@ class PublisherSettingsForm(forms.ModelForm):
             "allow_community_campaigns",
             "allow_house_campaigns",
             "record_placements",
+        ]
+
+
+class InviteUserForm(forms.ModelForm):
+
+    """Used to invite users to collaborate on an advertiser/publisher."""
+
+    def __init__(self, *args, **kwargs):
+        """Add the form helper and customize the look of the form."""
+        super().__init__(*args, **kwargs)
+        self.helper = FormHelper()
+        self.helper.layout = Layout(
+            Fieldset(
+                "",
+                Field("name"),
+                Field("email", placeholder="user@yourdomain.com"),
+                css_class="my-3",
+            ),
+            Submit("submit", "Send invite"),
+        )
+
+    def clean_email(self):
+        email = self.cleaned_data.get("email")
+        return User.objects.normalize_email(email)
+
+    def save(self, commit=True):
+        user = super().save(commit)
+        user.invite_user()
+        # You will need to add the user to the publisher/advertiser in the view
+        return user
+
+    class Meta:
+        model = User
+        fields = [
+            "name",
+            "email",
         ]
--- a/adserver/views.py
+++ b/adserver/views.py
@@ -964,12 +964,13 @@ class AdvertiserAuthorizedUsersView(
 
     """Authorized users for an advertiser."""
 
+    context_object_name = "users"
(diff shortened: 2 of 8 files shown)


adserver/views.py
```diff
@@ -956,6 +958,110 @@ def get_context_data(self, **kwargs):
...
+):
+
+    """Authorized users for an advertiser."""
+
+    model = get_user_model()
+    template_name = "adserver/advertiser/users.html"
+
+    def get_context_data(self, **kwargs):  # pylint: disable=arguments-differ
+        context = super().get_context_data(**kwargs)
+        context.update({"advertiser": self.advertiser, "users": self.object_list})
```
ericholscher, line 972: Any reason we're setting `users` instead of `context_object_name = users`? 

https://docs.djangoproject.com/en/3.2/topics/class-based-views/generic-display/#making-friendly-template-contexts

adserver/views.py
```diff
@@ -956,6 +958,110 @@ def get_context_data(self, **kwargs):
...
+        return super().dispatch(request, *args, **kwargs)
+
+    def form_valid(self, form):
+        result = super().form_valid(form)
+        self.object.advertisers.add(self.advertiser)
+        messages.success(
+            self.request,
+            _("Successfully invited %(user)s to %(advertiser)s")
+            % {"user": self.object.email, "advertiser": self.advertiser},
+        )
```
ericholscher, line 1005: Are there any error cases here? I guess most (eg. invalid email) would be handled in the form.

adserver/forms.py
```diff
@@ -392,3 +394,39 @@ class Meta:
             "allow_house_campaigns",
             "record_placements",
         ]
+
+
+class InviteUserForm(forms.ModelForm):
+
+    """Used to invite users to collaborate on an advertiser/publisher."""
```
ericholscher, line 401: Did we not build this for publishers yet?

adserver/views.py
```diff
@@ -956,6 +958,110 @@ def get_context_data(self, **kwargs):
...
+
+
+class AdvertiserAuthorizedUsersRemoveView(
+    AdvertiserAccessMixin, UserPassesTestMixin, TemplateView
+):
+
+    """
+    Remove authorized users for an advertiser.
+
+    This doesn't remove or deactivate the user - just removes them from the advertiser.
```
ericholscher, line 1026: I wonder what our UX looks like for users with no advertisers or publishers. Might be something to consider here, once a user gets removed it might just look confusing and broken.

adserver/views.py
```diff
@@ -956,6 +958,110 @@ def get_context_data(self, **kwargs):
...
+        self.user = get_object_or_404(
+            get_user_model(),
+            pk=self.kwargs["user_id"],
+            advertisers=self.advertiser,
+        )
+        return super().dispatch(request, *args, **kwargs)
+
+    def post(self, request, *args, **kwargs):
+        if self.user == self.request.user:
+            messages.error(self.request, _("You cannot remove your own access"))
```
ericholscher, line 1044: Should probably disable this in the frontend UI, if we haven't already.

ericholscher (approved):
Looks like a useful feature. 👍

adserver/forms.py
davidfischer: No. I didn't. I haven't gotten many requests to do that. Have you?

adserver/views.py
davidfischer: Sure.

adserver/views.py
davidfischer: They get a screen with "You don't have access to anything". https://github.com/readthedocs/ethical-ad-server/blob/main/adserver/templates/adserver/dashboard.html#L43-L45

adserver/forms.py
ericholscher: Nope, and if so I can always just add them in the admin.

davidfischer: My plan was to allow a view for a publisher to re-use the form. The view can't be re-used exactly although maybe refactored into a mixin.

adserver/views.py
davidfischer: Ya, the only error case is an invalid email.

adserver/views.py
davidfischer: I didn't know about that. That's awesome.

PR #374, opened 2021-04-29
Allow filtering advertiser reports by flight

davidfischer:
This just allows filtering the geo and pubisher reports (the publisher report for advertiser) by the flight. I got an advertiser asking for this.

### DIFF
--- a/adserver/views.py
+++ b/adserver/views.py
@@ -798,11 +798,17 @@ class AdvertiserGeoReportView(AdvertiserAccessMixin, GeoReportMixin, BaseReportV
         advertiser_slug = kwargs.get("advertiser_slug", "")
         advertiser = get_object_or_404(Advertiser, slug=advertiser_slug)
 
+        flight_slug = self.request.GET.get("flight", "")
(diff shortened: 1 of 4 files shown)


adserver/views.py
```diff
@@ -798,11 +798,17 @@ def get_context_data(self, **kwargs):
         advertiser_slug = kwargs.get("advertiser_slug", "")
         advertiser = get_object_or_404(Advertiser, slug=advertiser_slug)
 
+        flight_slug = self.request.GET.get("flight", "")
+        flight = Flight.objects.filter(
+            campaign__advertiser=advertiser, slug=flight_slug
+        ).first()
```
ericholscher, line 804: Does the rest of the code handle `flight=None` in a good way? Should we error on this? Would only happen w/ URL hacking I guess, so probably not a huge deal.

ericholscher (approved):

adserver/views.py
davidfischer: It handles `flight=None` as not filtering on flights. It will display results for all flights.

PR #375, opened 2021-05-04
Add geo & keyword staff reports

ericholscher:
This is a small report refactor to move all aggregate reports to "staff
reports".

A few things to note:

* All publisher/advertiser still use a custom template, since they do full
listings
* This unifies most of the logic for the All keyword/geo/uplift reports
* These reports can get slow when run against a large time frame

I also implemented breakdown data via keyword URL but didn't include it in
the filters, since it's a staff-only report (eg.
/all/report/geos/?country=us).


adserver/reports.py
```diff
@@ -343,4 +343,4 @@ class PublisherUpliftReport(PublisherReport):
 
     model = UpliftImpression
     index = "publisher.name"
-    order = "index"
+    order = "-views"
```
ericholscher, line 346: Sorts uplift by views instead of alphabetically.

adserver/templates/adserver/dashboard.html
```diff
@@ -14,14 +14,26 @@
...
+    <h2>{% trans 'Staff Reports' %}</h2>
+    <p>
+      {% trans 'Reports only visible to staff users.' %}
+    </p>
+    <ul>
+        <li><a href="{% url 'all_advertisers_report' %}">{% trans 'All advertiser report' %}</a></li>
+        <li><a href="{% url 'all_publishers_report' %}">{% trans 'All publisher report' %}</a></li>
+        <li><a href="{% url 'all_keyword_report' %}">{% trans 'All keyword report' %}</a></li>
+        <li><a href="{% url 'all_geo_report' %}">{% trans 'All geo report' %}</a></li>
+        <li><a href="{% url 'publisher_uplift_report' %}">{% trans 'All uplift report' %}</a></li>
```
davidfischer, line 27: I know the other reports have "all" in front of them but this one doesn't in the actual report template.

```suggestion
        <li><a href="{% url 'publisher_uplift_report' %}">{% trans 'Uplift report' %}</a></li>
```

adserver/templates/adserver/reports/base-all.html
```diff
@@ -1,58 +1,41 @@
...
       <option value="{{ option }}" {% if sort == option %}selected{% endif %}>{{ option }}</option>
       {% endfor %}
     </select>
   </div>
 {% endblock additional_filters %}
 
-
 {% block explainer %}
 <section class="mb-5">
   <h3>About this report</h3>
```
davidfischer, line 20:
```suggestion
  <h3>{% trans 'About this report' %}</h3>
```

adserver/templates/adserver/reports/base-all.html
```diff
@@ -1,58 +1,41 @@
...
-  <p>This report shows the additional advertising data that we can attribute to Acceptable Ads.</p>
   <p>
-    <em>This report shows all publishers and updates periodically. All previous days data is complete.</em>
+  <em>
+  This is a staff-only report. "Our Rev" reports 70% for all publishers in these views.
+  </em>
```
davidfischer, line 24:
```suggestion
  <em>{% trans 'This is a staff-only report. "Our Rev" assumes a 70% revenue share for all publishers in these views.' %}</em>
```

adserver/templates/adserver/reports/all-geos.html
```diff
@@ -0,0 +1,15 @@
+{% extends "adserver/reports/base-all.html" %}
+{% load humanize %}
+{% load i18n %}
+
+
+{% block title %}{% trans 'All Geo Report' %}{% endblock %}
```
davidfischer, line 6: Do you think it's better to name these the "Staff Geo Report" or something along those lines? This "all X" sort of breaks down for uplift.

davidfischer (approved):
This change looks like an improvement. I loaded it up and clicked around and everything seemed right. We may have to sort of "test in production" with these. It's nice that there's no migrations.

I do have a few questions:

* Is there a reason we're forcing a 70% revshare on the calculations? Are they too expensive to be done? They should be able to be done in the database.
* In general the breakdown of Staff Reports, Publishers, and Advertisers is a good improvement. Should we call them "Staff X report" instead of "All X report" though?

adserver/templates/adserver/reports/all-geos.html
ericholscher: Yea, this is better naming.

davidfischer:
New changes look :+1:

PR #376, opened 2021-05-05
Make report queries faster

ericholscher:
This only gets the data we need from the DB,
which makes this query much faster locally.
Not sure how much faster it will be in prod,
but should help some.

### DIFF
--- a/adserver/reports.py
+++ b/adserver/reports.py
@@ -193,7 +193,15 @@ class PublisherReport(BaseReport):
 
         results = {}
 
-        for impression in queryset:
+        for impression in queryset.only(
+            "advertisement__flight",
+            "publisher__revenue_share_percentage",
+            "decisions",
+            "offers",
+            "views",
+            "clicks",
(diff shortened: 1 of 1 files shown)


adserver/reports.py
```diff
@@ -193,7 +193,15 @@ def generate(self):
...
 
-        for impression in queryset:
+        for impression in queryset.only(
+            "advertisement__flight",
+            "publisher__revenue_share_percentage",
+            "decisions",
+            "offers",
+            "views",
+            "clicks",
+            self.index,
```
ericholscher, line 203: This needs a bit more logic to work properly -- but I think this change is good with a bit of tuning.

davidfischer: Ya, I think we might need to translate dot notation to dunder notation or something.

davidfischer (commented):
I was skeptical that this would actually improve performance much but maybe it will. Sometimes these things just have to be tested empirically.

PR #380, opened 2021-05-12
I missed this when adding CTR to advertiser reports

davidfischer:


MxRaCohen (approved):
Looks good. Perhaps add _specified_ report period

PR #381, opened 2021-05-14
Setup a support form

davidfischer:
- Sends email by default
- Can be hooked up to an external help desk (eg. Front)
- Requires a login but captures the user info from the user submitting the support request.


adserver/templates/adserver/includes/publisher-paid-notice.html
```diff
@@ -2,5 +2,6 @@
 
 
 {% if publisher and not publisher.allow_paid_campaigns %}
-<div class="alert alert-warning" role="alert">{% trans 'Your publisher account is not approved for paid campaigns. When you are ready for us to review your integration, please contact us.' %}</div>
+{% url 'support' as support_url %}
```
ericholscher, line 5: I was noticing that this warning wasn't included on all pages. Should it be part of the base template instead of an include?

config/settings/base.py
```diff
@@ -370,5 +370,10 @@
 ADSERVER_HTTPS = False  # Should be True in most production setups
 ADSERVER_STICKY_DECISION_DURATION = 0
 
+# For customer support emails
+ADSERVER_SUPPORT_FROM_EMAIL = env("ADSERVER_SUPPORT_FROM_EMAIL", default=SERVER_EMAIL)
```
ericholscher, line 374: No doc for this one?

adserver/templates/adserver/base.html
```diff
@@ -28,11 +28,17 @@
...
             {% endif %}
+
             <a class="dropdown-item" href="{% url 'api_token_list' %}">
               <span class="fa fa-cog fa-fw mr-2 text-muted" aria-hidden="true"></span>
               <span>{% trans 'API token' %}</span>
             </a>
             <div class="dropdown-divider"></div>
+            <a class="dropdown-item" href="{% url 'support' %}">
+              <span class="fa fa-envelope fa-fw mr-2 text-muted" aria-hidden="true"></span>
+              <span>{% trans 'Support' %}</span>
```
ericholscher, line 39: Should probably put it in the footer as well.

ericholscher (approved):
Useful change 👍

config/settings/base.py
davidfischer: I'll remove it

davidfischer: It actually wasn't even used. It always used [DEFAULT_FROM_EMAIL](https://docs.djangoproject.com/en/2.2/ref/settings/#std:setting-DEFAULT_FROM_EMAIL) which seems correct.

adserver/templates/adserver/includes/publisher-paid-notice.html
davidfischer: Sure, that works.

davidfischer:
I did the feedback changes and prepared the ops changes for new envvars.

ericholscher:
Let's ship it next week!

PR #386, opened 2021-05-26
Null out old client IDs

davidfischer:
ClientIDs are used for fraud prevention and for the ad pinning that shows the same ad to the same person for five seconds. These IDs are a 1-way hash made up of the user agent, the IP address, a salt, and our secret key. They aren't exposed to end users. They change over time as user agent strings or IPs change but they are steady over very short time periods and let us analyze things from a fraud perspective.

This change changes a few things:

- The IP will be anonymized before it is hashed. This will mean that folks with the same UA and similar IPs (last 2 bytes are dropped) will get the same ClientID. For most real fraud, this should be sufficient.
- After 90 days, Client IDs will be nulled out in the database.

**Performance note:** If this is approved, the performance on the live database will probably need to be assessed as nulling out parts of this table could be very expensive. It may need to be done in chunks to be performant.

### DIFF
--- a/adserver/models.py
+++ b/adserver/models.py
@@ -1462,10 +1462,11 @@ class AdBase(TimeStampedModel, IndestructibleModel):
     )
 
     # User Data
-    ip = models.GenericIPAddressField(_("Ip Address"))  # should be anonymized
+    ip = models.GenericIPAddressField(_("Ip Address"))  # anonymized
     user_agent = models.CharField(
         _("User Agent"), max_length=1000, blank=True, null=True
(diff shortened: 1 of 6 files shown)


davidfischer:
> We should perhaps babysit it & test it manually prior to merging this and deploying it to autorun.

I tried it and it's bad. I'm going to adjust it to do this 1000 at a time.

davidfischer:
I changed the code to something more performant. I tested it in prod and it took about 2s per 1000.

PR #387, opened 2021-05-27
Move the add advertiser interface to a staff action

davidfischer:
- Instead of a CLI, there's now a web interface
- Basically duplicates the logic of the CLI. Additional logic like customizing the created flight could be added on

There's a few open questions:

- Should we move the staff reports here?
- Is a left nav the right place for this?

## Screenshot
Screenshot from 2021-05-26 22-06-25


adserver/staff/forms.py
```diff
@@ -0,0 +1,171 @@
...
+class CreateAdvertiserForm(forms.Form):
+
+    """
+    Creates an advertiser.
+
+    This isn't a simple model form because it creates a few additional objects as well
+    such as a user (needed to create a stripe account), campaign, and initial flight.
+    """
+
+    DEFAULT_CPM = 3.33
```
ericholscher, line 33:
```suggestion
    # TODO: Make these configurable in the web UI as a dropdown of common values. 
     DEFAULT_CPM = 3.33
```

adserver/staff/forms.py
```diff
@@ -0,0 +1,171 @@
...
+            targeting_parameters={
+                "include_countries": self.DEFAULT_COUNTRY_TARGETING,
+            },
+        )
+
+        return advertiser
+
+    def create_stripe_customer(self, user, advertiser):
+        """Setup a Stripe customer for this user."""
+        if not settings.STRIPE_SECRET_KEY:
```
ericholscher, line 148: Is it worth logging in this case? Probably not, but not sure.

ericholscher (approved):
Simple enough 👍 -- Excited to get more of this out of the CLI.

adserver/staff/forms.py
davidfischer: I don't think it's necessary. This will be logged in a number of places since a few other things fail if the Stripe key isn't set.

PR #388, opened 2021-06-04
Add RegionTopic index modeling

ericholscher:
This doesn't create the views on the data yet,
only the initial modeling for review.


adserver/tasks.py
```diff
@@ -218,6 +232,128 @@ def daily_update_keywords(day=None):
...
+            flight_targeting = json.loads(
+                values["advertisement__flight__targeting_parameters"]
+            )
+            flight_keywords = set(flight_targeting.get("include_keywords", {}))
+
+            matched_keywords = publisher_keywords & flight_keywords
+
+            topics = set()
+            for keyword in matched_keywords:
+                if keyword in data_science:
```
ericholscher, line 294: This is definitely a bit hacky, but I didn't want to change the data model too heavily. Could definitely be refactored to be a bit nicer using a dict or similar.

adserver/tasks.py
```diff
@@ -218,6 +232,128 @@ def daily_update_keywords(day=None):
...
+
+            if not matched_keywords:
+                # No matched keywords
+                for keyword in publisher_keywords:
+                    if keyword in no_keywords:
+                        no_keywords[keyword] += 1
+                    else:
+                        no_keywords[keyword] = 1
+
+            if country in us_ca:
```
davidfischer, line 326: Similar to topics, there will be overlap between regions. I think we need to have a regions set as well.

adserver/tasks.py
```diff
@@ -218,6 +232,128 @@ def daily_update_keywords(day=None):
...
+                "keywords",
+                "country",
+                "advertisement__flight__targeting_parameters",
+                "total",
+            )
+        ):
+            if not (
+                values["keywords"]
+                and values["country"]
+                and values["advertisement__flight__targeting_parameters"]
```
davidfischer, line 277: Are there a lot of flights that don't have targeting?

adserver/tasks.py
```diff
@@ -218,6 +232,128 @@ def daily_update_keywords(day=None):
...
+            keywords = json.loads(values["keywords"])
+            country = values["country"]
+            publisher_keywords = set(keywords)
+
+            flight_targeting = json.loads(
+                values["advertisement__flight__targeting_parameters"]
+            )
+            flight_keywords = set(flight_targeting.get("include_keywords", {}))
+
+            matched_keywords = publisher_keywords & flight_keywords
```
davidfischer, line 290: Don't we want matched as well as unmatched keywords? Like just because a geo-targeted only ad is shown, we want to know if this could have been a data science ad.

davidfischer (commented):
This looks pretty close. I do want to make sure we're indexing even when there's overlap between topics and/or regions though.

adserver/tasks.py
ericholscher: I was assuming that region was exclusive. What is the case for overlapping regions?

adserver/tasks.py
ericholscher: Not sure. Probably mostly house ads.

adserver/tasks.py
ericholscher: Do we? That will pretty quickly blow up the index if we're indexing every keyword on the publisher.

ericholscher: Thinking more, I think you're right. We need this to get a real fill rate.

ericholscher: What should we be doing with the unmatched keywords though? That is basically "keywords a publisher had, but didn't get an ad served against". It presumably shouldn't be part of the model record attached to the ad (since the ad wasn't served against that keyword) -- so should it have a blank advertisement?

adserver/tasks.py
davidfischer: Our default targeting (NA/EU/Aus/NZ) includes Australia and New Zealand. However, they are also part of APAC. There are lots of semi-overlapping regions.

adserver/tasks.py
davidfischer: Basically what we want to know is how many `$TOPIC` ads we *could have* served in `$REGION`. If we got a matching ad request for `$TOPIC`/`$REGION` and chose to serve a more generic ad isn't relevant.

adserver/tasks.py
ericholscher: Sure, but we don't have any overlapping regions in our data, so I'm not sure how I'd map something to multiple regions with the current data.

ericholscher:
This should now be ready for review.

adserver/admin.py
```diff
@@ -856,19 +857,37 @@ class PublisherGroupAdmin(admin.ModelAdmin):
...
-admin.site.register(GeoImpression, GeoImpressionAdmin)
-admin.site.register(PlacementImpression, PlacementImpressionAdmin)
-admin.site.register(KeywordImpression, KeywordImpressionAdmin)
 admin.site.register(AdType, AdTypeAdmin)
 admin.site.register(Advertisement, AdvertisementAdmin)
 admin.site.register(Flight, FlightAdmin)
 admin.site.register(Campaign, CampaignAdmin)
+
+# Don't register Impression Admin's outside dev, since they will just 502 from too much data.
+if settings.DEBUG:
```
ericholscher, line 887: Think this is what we want, but worth noting.

adserver/tasks.py
```diff
@@ -218,6 +233,105 @@ def daily_update_keywords(day=None):
...
+                elif keyword in devops:
+                    topic = "devops"
+                elif keyword in python:
+                    topic = "python"
+                elif keyword in blockchain:
+                    topic = "blockchain"
+                elif keyword in game_dev:
+                    topic = "game-dev"
+                else:
+                    continue
```
ericholscher, line 297: We are _not_ counting views that have no keywords here. I wasn't sure if we wanted an `other` topic. I'm guessing it will be most views (because it would include every junk keyword as `other`)

adserver/tasks.py
ericholscher: I feel like we should probably keep this contract on the regions:

* Each country maps to exactly one subregion
* We can combine subregions into larger regions if we want (eg. Western Europe & Eastern Europe = Europe) -- but nothing would be tagged `Europe` in our data, just in aggregations.

adserver/admin.py
davidfischer: Do they all 502? Regardless, I'm OK with it.

adserver/tasks.py
davidfischer: Ok.

adserver/tasks.py
davidfischer: I think that's OK actually. However, an `other` might be decent too.

davidfischer (approved):

davidfischer:
I fixed up the migration order and fixed the merge issue.

ericholscher:
Thanks! I'll update my other PR with these migraitons.

PR #394, opened 2021-06-10
Add initial Staff Payouts view

ericholscher:
This will be used for processing payouts,
and will replace ``./manage.py payout`` which is the current implementation.

This is just an initial list view,
and doesn't fully port all the logic for sending email, etc.
That will come in a later PR.


adserver/staff/views.py
```diff
@@ -40,3 +43,70 @@ def get_success_url(self):
...
+    * publisher: Filter to a specific publisher
+    * limit: Show only up to ``limit`` publishers, mostly for testing & debugging
+    """
+
+    template_name = "adserver/staff/publisher-payout.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        limit = self.request.GET.get("limit", 50)
```
davidfischer, line 63: Just a note that this can fail (down below where used) if a non-int is passed. Pretty minor.

davidfischer (approved):

adserver/staff/forms.py
```diff
@@ -170,3 +176,64 @@ def save(self):
...
+            "sender_name": self.cleaned_data["sender"],
+            "subject": self.cleaned_data["subject"],
+            "options": {"archive": self.cleaned_data["archive"]},
+            "body": self.cleaned_data["body"],
+        }
+        if author:
+            payload["author_id"] = author
+        url = f"https://api2.frontapp.com/channels/{channel}/messages"
+        log.info("Sending email")
+        requests.request("POST", url, json=payload, headers=headers)
```
davidfischer, line 225: Maybe not in this PR but if we decide we want to send front emails from a few places, we could create a Front [email backend](https://docs.djangoproject.com/en/2.2/topics/email/#topic-email-backends).

adserver/templates/adserver/staff/publisher-payout-list.html
```diff
@@ -0,0 +1,58 @@
...
+            {% if payout_data.payout and payout_data.payout.status == "paid" %}
+            No action required
+            {% elif payout_data.payout and payout_data.payout.status == "emailed" %}
+            <a href="{% url "staff-finish-publisher-payout" publisher.slug %}">{% trans "Finish Payout" %}</a>
+            {% else %}
+            <a href="{% url "staff-start-publisher-payout" publisher.slug %}">{% trans "Start Payout" %}</a>
+            {% endif %}
+            </td>
+          </tr>
+        {% endfor %}
```
davidfischer, line 54:
```suggestion
        {% empty %}
          <td colspan="7">{% trans 'No pending payouts' %}</td>
        {% endfor %}
```

davidfischer (approved):
Looks good. I didn't fully test it end-to-end but it looks like it should work.

PR #395, opened 2021-06-14
Updates the weighting algorithm

davidfischer:
- Periodically calculates the CTR for publishers for the past week
- Uses estimated CTR in the weighting algorithm for selecting an ad, but only in CPC ads
- Effectively uses "estimated CPM" as a weighting factor when choosing ads

Essentially, this will bias our ad selection toward ads that will make the most money for us and our publishers. For example, if a publisher sees an average CTR of 0.3% (~3x average) they will be more likely to receive CPC ads since those are more likely to make more money.

**NOTE:** Because this ads a migration, it conflicts in migration numbering with #388. Be careful when merging.


adserver/decisionengine/backends.py
```diff
@@ -325,7 +325,9 @@ def select_flight(self):
                     )
                 ):
                     # NOTE: takes into account views for CPM ads
```
ericholscher, line 327: Should we update this comment to note CPC?

adserver/models.py
```diff
@@ -631,33 +639,42 @@ def clicks_needed_today(self):
...
         # This is naive but we are counting a click as being worth 1,000 views
         impressions_needed += math.ceil(self.views_needed_today() / 1000.0)
         impressions_needed += self.clicks_needed_today()
 
-        # Cost should be a float from around .5-5, averaging 2
-        cost = float(self.cpm or self.cpc)
+        # Use the publisher CTR if available
+        # Otherwise, use this flight's average CTR
+        estimated_ctr = float(self.ctr())
+        if publisher and publisher.sampled_ctr > 0.01:
```
ericholscher, line 660: If this is just checking against `0.0` I think the `if` check will work. Might also just do `!= 0.0` to be more explicit?

adserver/models.py
```diff
@@ -631,33 +639,42 @@ def clicks_needed_today(self):
...
 
         # This is naive but we are counting a click as being worth 1,000 views
         impressions_needed += math.ceil(self.views_needed_today() / 1000.0)
         impressions_needed += self.clicks_needed_today()
 
-        # Cost should be a float from around .5-5, averaging 2
-        cost = float(self.cpm or self.cpc)
+        # Use the publisher CTR if available
+        # Otherwise, use this flight's average CTR
+        estimated_ctr = float(self.ctr())
```
ericholscher, line 659: Seems this block should be in the `if self.cpc` block, since it's only used in that case, and makes the logic clearer.

I might even suggest an `if self.cpc` and `elif self.cpm` blocks to make it really obvious what we're doing.

adserver/tasks.py
```diff
@@ -281,3 +284,21 @@ def remove_old_client_ids(days=90):
...
+@app.task()
+def calculate_publisher_ctrs(days=7):
+    """Calculate average CTRs for paid ads on a publisher for the last X days."""
+    sample_cutoff = get_ad_day() - datetime.timedelta(days=days)
+
+    for publisher in Publisher.objects.all():
+        queryset = AdImpression.objects.filter(
+            date__gte=sample_cutoff,
+            publisher=publisher,
+            advertisement__flight__campaign__campaign_type=PAID_CAMPAIGN,
```
ericholscher, line 298: 👍

adserver/tasks.py
```diff
@@ -281,3 +284,21 @@ def remove_old_client_ids(days=90):
...
+        )
+        report = PublisherReport(queryset)
+        report.generate()
```
ericholscher, line 301: There's probably a more efficient way to do this with a query, but this is fine and incentivizes us to make report generation as efficient as possible 👍

ericholscher (approved):
Exciting!

adserver/models.py
davidfischer: I'm always really scared to do precise comparisons against floats due lack of precision.

adserver/models.py
```diff
@@ -631,33 +639,44 @@ def clicks_needed_today(self):
...
         """
         impressions_needed = 0
 
         # This is naive but we are counting a click as being worth 1,000 views
         impressions_needed += math.ceil(self.views_needed_today() / 1000.0)
         impressions_needed += self.clicks_needed_today()
 
-        # Cost should be a float from around .5-5, averaging 2
-        cost = float(self.cpm or self.cpc)
+        if self.cpc:
```
ericholscher, line 657: Much cleaner 👍

ericholscher (approved):
Looks good! Let's get it shipped 🎉

PR #396, opened 2021-06-14
This process is consuming the server

davidfischer:
I would still like to figure out a solution here but this is consuming the server. I think we need to archive offers on a daily basis automatically and that might shrink the table to reasonable sizes. After that, this might make sense.


ericholscher (approved):
Ah, I thought we'd already made a plan for not having this kill the DB. Definitely shouldn't run this against a huge table regularly.

davidfischer:
Actually the queries do run reasonably quickly -- just ~2s per 1000. The problem is that there's just too much.

PR #399, opened 2021-06-15
Track historical changes to select models

davidfischer:
- Uses [django simple history](https://github.com/jazzband/django-simple-history) to track model changes. This had a few advantages over competitors for us. It does allow reverting if needed. You can programmatically query the history if you want to [diff](https://django-simple-history.readthedocs.io/en/latest/history_diffing.html) things. One drawback is it does create a table for historical models for each model you want to track.
- I added change tracking to all the core ad models and to the user model but not any of the reporting models (Offers, AdImpressions, etc.). The latter would just store too much data.
- I verified that viewing and clicking an ad doesn't store anything to these new tables.
- Custom [change messages](https://django-simple-history.readthedocs.io/en/latest/historical_model.html#change-reason) can be added when adding/updating in order to track why a change was made. I added a few of these for adding authorized users, doing a payout, and creating an advertiser. The user making the change is captured by middleware.
- When deploying this, if we want to capture the current state of the models (eg. before we've made any changes), we need to run `./manage.py populate_history --auto`. See ["existing projects"](https://django-simple-history.readthedocs.io/en/latest/quick_start.html#existing-projects)


ericholscher (approved):
Looks pretty simple -- I didn't do a deep dive here, but if it does what it says on the tin then I think we're good 👍

PR #400, opened 2021-06-16
Clean up a number of payout workflow issues

ericholscher:
We hit a bunch of things during payout that slowed us down.
This resolves a few of them:

* Easier to compare payout to last month
* Redirect Start form submission -> Finish form submission.
* Day of week in the date on reports
* Note publisher & email on Payout start & finish page

### DIFF
--- a/adserver/reports.py
+++ b/adserver/reports.py
@@ -286,7 +286,8 @@ class PublisherReport(BaseReport):
     def get_index_display(self, index):
         """Handle making sure dates use the same 3 letter syntax as Django."""
         if isinstance(index, datetime.date):
-            return index.strftime("%b %d, %Y")
+            # %a is the day of week, so you know which are weekends
+            return index.strftime("%b %d, %Y (%a)")
 
         return super().get_index_display(index)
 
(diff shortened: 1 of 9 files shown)


adserver/reports.py
```diff
@@ -286,7 +286,8 @@ def calculate_totals(self):
     def get_index_display(self, index):
         """Handle making sure dates use the same 3 letter syntax as Django."""
         if isinstance(index, datetime.date):
-            return index.strftime("%b %d, %Y")
+            # %a is the day of week, so you know which are weekends
+            return index.strftime("%b %d, %Y (%a)")
```
davidfischer, line 290: Just a note that this cascades down all the way into the CSV exports. It also doesn't apply to advertisers (only publishers). I'm mostly OK with it but just wanted to air these notes.

davidfischer (approved):

adserver/reports.py
ericholscher: Yea -- not sure the best way to handle it. I think changing the CSS on the weekends is probably the best way to handle this (make them a slightly diff background color) -- but this seemed most explicit for now.

PR #405, opened 2021-06-23
Add Staff Add Publisher View

ericholscher:
This will allow all staff users to add a publisher in the web UI.
It basically just ports the logic from the existing management command,
using the same structure as the Create Advertiser view.


adserver/staff/forms.py
```diff
@@ -187,6 +188,104 @@ def save(self):
...
+
+    """
+    Creates a publisher.
+
+    This isn't a simple model form because it creates a few additional objects as well
+    """
+
+    DEFAULT_GROUP = "ethicalads-network"
+
+    # Advertiser information
```
davidfischer, line 201:
```suggestion
    # Publisher information - this isn't a URLField because we just want the domain
```

adserver/staff/forms.py
```diff
@@ -187,6 +188,104 @@ def save(self):
...
+    site = forms.CharField(label=_("Publisher site"), max_length=200)
+
+    # User information
+    user_name = forms.CharField(label=_("User Name"), max_length=200)
+    user_email = forms.EmailField(label=_("User Email"))
+
+    # Publisher Info
+    keywords = forms.CharField(
```
davidfischer, line 209: Should we move this up under the publisher info section above? Also, we probably need some sort of validator here, right?

davidfischer (approved):

adserver/staff/forms.py
ericholscher: Yea, probably. I don't think we currently have one in the admin, but seems like a good plan. We're just parsing it as a CSV basically, so should we just ensure it's ASCII & ``?

davidfischer: I guess it's a staff form so maybe it's OK as-is. Maybe just ensure it's lowercase and contains only `[a-z0-9-]+`

ericholscher: Cool -- I will add that in the future, but think this would be good to unblock Ra in the meantime.

PR #407, opened 2021-06-29
Send Slack notifications on completed flights

davidfischer:
- This PR refactored the `generate_absolute_url` function to accept a path instead of a view and kwargs.
- We will need to generate and set a SLACK_TOKEN for this to work in production. It can be set as an environment variable.
- Relies on [django-slack](https://django-slack.readthedocs.io/)

### DIFF
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -5,6 +5,7 @@ import logging
 from collections import defaultdict
 
 from celery.utils.iso8601 import parse_iso8601
+from django.conf import settings
 from django.db.models import Count
 from django.db.models import F
 from django.utils.timezone import is_naive
@@ -438,10 +439,11 @@ def notify_of_completed_flights():
             log.info("Flight %s finished in the last day.", flight)
 
             # Send notification about this flight
-            slack_message(
-                "adserver/slack/flight-complete.slack",
-                {
-                    "flight": flight,
-                    "flight_url": generate_absolute_url(flight.get_absolute_url()),
-                },
-            )
+            if settings.SLACK_TOKEN:
+                slack_message(
+                    "adserver/slack/flight-complete.slack",
+                    {
+                        "flight": flight,
+                        "flight_url": generate_absolute_url(flight.get_absolute_url()),
+                    },
+                )
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -9,6 +9,7 @@ from django.db.models import Count
 from django.db.models import F
 from django.utils.timezone import is_naive
 from django.utils.timezone import utc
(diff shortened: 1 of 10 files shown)


adserver/tasks.py
```diff
@@ -417,3 +420,28 @@ def calculate_publisher_ctrs(days=7):
...
+    """Send a notification about flights which completed in the last day."""
+    cutoff = get_ad_day() - datetime.timedelta(days=1)
+    for flight in Flight.objects.filter(live=True).filter():
+        if (
+            flight.clicks_remaining() == 0
+            and flight.views_remaining() == 0
+            and AdImpression.objects.filter(
+                date__gte=cutoff,
+                advertisement__flight=flight,
+            ).exists()
```
ericholscher, line 436: This is good logic, but I wonder if we should add some state to make sure this happens? If we're confident this will run daily and complete thats good, but if it fails we won't catch it. 

This is fine for a v1, but I think we probably want something on the model defining `alerted` or similar, especially if we're emailing users.

adserver/templates/adserver/slack/flight-complete.slack
```diff
@@ -0,0 +1,6 @@
+{% extends django_slack %}
+
+
+{% block text %}
+Flight {{ flight }} for {{ flight.campaign.advertiser }} wrapped up today: {{ flight_url }}
```
ericholscher, line 5: Should we include any data about it? CTR? Spend? Other active flights?

adserver/tasks.py
```diff
@@ -417,3 +420,28 @@ def calculate_publisher_ctrs(days=7):
 
         publisher.sampled_ctr = report.total["ctr"]
         publisher.save()
+
+
+@app.task()
+def notify_of_completed_flights():
```
ericholscher, line 426: Should this check for a `SLACK_TOKEN`?

ericholscher (approved):
Solid improvement! One step closer to emailing folks 🎉

adserver/tasks.py
davidfischer: I agree completely. If we're emailing users, we need an additional flag.

adserver/tasks.py
davidfischer: I originally had that and I took it out but maybe we should have it (in the inner loop).

adserver/templates/adserver/slack/flight-complete.slack
davidfischer: I'm not sure what the most relevant pieces are although those are pretty good. I figure get a v1 out and iterate.

PR #409, opened 2021-07-08
Initial staff interface for flight targeting and size updates

davidfischer:
- Has hotlinks for setting standard targeting
- Uses same model validation
- Currently a staff-only interface but might be opened up later

## Screenshot
Screenshot from 2021-07-08 15-46-58

### DIFF
--- a/adserver/forms.py
+++ b/adserver/forms.py
@@ -31,13 +31,40 @@ from .models import AdType
 from .models import Advertisement
 from .models import Flight
 from .models import Publisher
+from .regiontopics import africa
+from .regiontopics import backend_web
+from .regiontopics import data_science
+from .regiontopics import devops
+from .regiontopics import eu_aus_nz
+from .regiontopics import exclude
+from .regiontopics import frontend_web
+from .regiontopics import latin_america
+from .regiontopics import python
+from .regiontopics import security_privacy
+from .regiontopics import us_ca
+from .regiontopics import wider_apac
+from .validators import TargetingParametersValidator
 
 
 log = logging.getLogger(__name__)  # noqa
 User = get_user_model()
 
 
-class FlightAdminForm(forms.ModelForm):
+class FlightMixin:
+
+    """Ensure the flight can't have both CPC and CPM."""
+
+    def clean(self):
+        cleaned_data = super().clean()
+        cpc = cleaned_data.get("cpc")
+        cpm = cleaned_data.get("cpm")
+        if cpc > 0 and cpm > 0:
+            raise forms.ValidationError(_("A flight cannot have both CPC & CPM"))
+
+        return cleaned_data
+
+
+class FlightAdminForm(FlightMixin, forms.ModelForm):
 
     """The form for flights used by the Django Admin."""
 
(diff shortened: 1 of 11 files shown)


adserver/forms.py
```diff
@@ -31,13 +31,40 @@
...
-class FlightAdminForm(forms.ModelForm):
+class FlightMixin:
+
+    """Ensure the flight can't have both CPC and CPM."""
+
+    def clean(self):
+        cleaned_data = super().clean()
+        cpc = cleaned_data.get("cpc")
+        cpm = cleaned_data.get("cpm")
+        if cpc > 0 and cpm > 0:
```
ericholscher, line 61: I've wondered about putting the value of the buy into this instead of the number of impressions. I feel like whenever I have to setup a flight, I work backwards from the value, but maybe thats ok. Seems like that's the real number people care about tho -- just a random thought I had the other day.

adserver/forms.py
```diff
@@ -60,13 +87,231 @@ class Meta:
...
+                    css_class="form-row",
+                ),
+                css_class="my-3",
+            ),
+            Fieldset(
+                _("Flight targeting"),
+                Div(
+                    "include_countries",
+                    Div(
+                        HTML(
```
ericholscher, line 162: I was trying to do this with crispy forms the other day (render HTML) -- interesting.

adserver/forms.py
```diff
@@ -60,13 +87,231 @@ class Meta:
...
+                                    <li class="list-inline-item"><a class="ea-update-field" href="#" data-value="{}" data-target-field="#id_include_countries">APAC</a></li>
+                                    <li class="list-inline-item"><a class="ea-update-field" href="#" data-value="{}" data-target-field="#id_include_countries">Latin America</a></li>
+                                    <li class="list-inline-item"><a class="ea-update-field" href="#" data-value="{}" data-target-field="#id_include_countries">Africa</a></li>
+                                </ul>
+                            """,
+                                ", ".join(us_ca),
+                                ", ".join(eu_aus_nz),
+                                ", ".join(wider_apac),
+                                ", ".join(latin_america),
+                                ", ".join(africa),
```
ericholscher, line 177: This feels a bit weird -- not really sure waht this is doing.

adserver/views.py
```diff
@@ -258,6 +260,47 @@ def get_object(self, queryset=None):  # pylint: disable=unused-argument
         )
 
 
+class FlightUpdateView(StaffAccessMixin, UserPassesTestMixin, UpdateView):
```
ericholscher, line 263: Should the StaffAccessMixin inhereit from the UserPassesTestMixin to make sure we don't miss that part when re-using the code?

ericholscher (approved):
Looks good 👍

adserver/forms.py
davidfischer: I actually think that is probably the right way.

adserver/forms.py
davidfischer: This takes the countries and puts them into `data-value` using `format_html` for safety. Then when somebody clicks the link "US/CA" it populates the relevant countries.

ericholscher: Oh I see, they are being passed into a string format?

adserver/views.py
davidfischer: Not only should it, but we already had a version of this mixin. We just need to define and use it in one place.

adserver/forms.py
davidfischer: Yes, but passed in in a way that they would be escaped if there was HTML in there.

PR #410, opened 2021-07-09
Notify when publisher metrics change

davidfischer:
- Sends a slack notification (if slack is configured)
- Notifies if a publisher's CTR or views change by 25% wk/wk

### DIFF
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -447,3 +447,63 @@ def notify_of_completed_flights():
                         "flight_url": generate_absolute_url(flight.get_absolute_url()),
                     },
                 )
+
+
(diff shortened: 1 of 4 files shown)


ericholscher (approved):
This looks good. Makes me want a read replica once again -- I'd like to prioritize that in the near future, though perhaps migrate to AWS first?

davidfischer:
> I'd like to prioritize that in the near future, though perhaps migrate to AWS first?

Compared with the data dumps, this should be pretty quick. I wouldn't worry too much about this specifically. However, yes, we should get a replica. Let's move to AWS first so we only have to do it once.

PR #411, opened 2021-07-12
Add some payout optimizations to make it faster

ericholscher:
We had some pageload slowness last time we did payouts.
This adds a bit of caching to the payout data,
and does a bit less work that we don't need on this page.

### DIFF
--- a/adserver/staff/views.py
+++ b/adserver/staff/views.py
@@ -4,6 +4,7 @@ import logging
 from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth.mixins import UserPassesTestMixin
+from django.core.cache import cache
 from django.http import Http404
 from django.shortcuts import get_object_or_404
 from django.shortcuts import redirect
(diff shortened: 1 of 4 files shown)


adserver/staff/views.py
```diff
@@ -91,28 +92,39 @@ class PublisherPayoutView(StaffUserMixin, TemplateView):
...
 
-        for publisher in queryset[:limit]:
-            data = generate_publisher_payout_data(publisher)
+        for publisher in queryset:
+            data = generate_publisher_payout_data(
+                publisher, include_current_report=False
+            )
+
+            # Cache payout data to make running payouts faster
+            data = cache.get(f"payout-{publisher.pk}")
```
davidfischer, line 120: If we're just going to cache all the data, couldn't we just cache the whole view?

davidfischer (approved):

adserver/staff/views.py
ericholscher: We need to have the publisher payout state change over time as we update it. We could in theory do that in the template, but that seems pretty hacky.

PR #412, opened 2021-07-13
Publisher (group) specific ad types

davidfischer:
- Allows ad types to be specific to a set of publishers (eg. RTD specific ads). They can still apply to all publishers too.
- If a campaign is not configured to target that publisher group, the option to configure ads of that type won't be presented
- Ad types can be complimentary or mutually exclusive. For example, the same ad can be presented without an image or with an image. But you can have two ad types that have incompatibly sized images for example.

## Rollout
After rolling this out, we could set the EA specific ads (image+text, text-only) as EA specific and the RTD ad formats (sidebar, deprecated ones) as RTD specific. There's no rush to do this and things will function after the migration without a problem. For almost all advertisers, this will make no difference but a few advertisers have campaigns that are RTD or EA specific. For those, the interface might be simplified slightly.


adserver/forms.py
```diff
@@ -203,9 +202,10 @@ class AdvertisementForm(AdvertisementFormMixin, forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         """Add the form helper and customize the look of the form."""
-        self.flight = None
         if "flight" in kwargs:
             self.flight = kwargs.pop("flight")
+        else:
+            raise RuntimeError("'flight' is required for the ad form")
```
ericholscher, line 208: 👍

adserver/models.py
```diff
@@ -365,6 +365,18 @@ def __str__(self):
     def ad_count(self):
         return Advertisement.objects.filter(flight__campaign=self).count()
 
+    def allowed_ad_types(self, exclude_deprecated=False):
+        """Get the valid ad types for this campaign."""
+        queryset = AdType.objects.filter(
+            models.Q(publisher_groups=None)
+            | models.Q(  # Global ad types across all publishers
```
ericholscher, line 372: Isn't this comment talking about `publisher_groups=None`?

adserver/models.py
```diff
@@ -802,6 +814,16 @@ class AdType(TimeStampedModel, models.Model):
         ),
     )
 
+    publisher_groups = models.ManyToManyField(
+        PublisherGroup,
+        related_name="adtypes",
+        blank=True,
```
ericholscher, line 820: m2m's default to null=True right? (eg. No objects)

ericholscher (approved):
This looks good 👍

adserver/models.py
davidfischer: Yes

adserver/models.py
davidfischer: This is awkward. This actually does apply above. I think it got formatted like this. I'll see if I can re-format it.

PR #416, opened 2021-07-20
Add noopener to external links

davidfischer:
This fixes the external links in the text of the ad. The other links (eg. "Ad by EthicalAds") is in the ad template and was changed in the template.

Fixes https://github.com/readthedocs/ethical-ad-client/issues/75


ericholscher (approved):
✨

Issue #359, opened 2021-04-07
Capture "view time" for ads


davidfischer:
This issue is to provide scoping around a new feature to capture the "view time" for a particular ad when it's viewed. This would require the server to store how long an ad has been viewed and the client to send updates on how long an ad was viewed for. We think that advertisers may find this valuable and might even pay more when ads are viewed for longer times (static ads providing revenue similar to video ads). Likewise, publishers with more "view time" per view should be compensated more than just a straight CPM number. This might also enable a change to where ads change periodically (eg. https://github.com/readthedocs/ethical-ad-client/issues/24) for SPAs or long-viewed pages.

**Changes to the server**

- The server would need a place to record the view time for a particular "view". This could probably just be a positive integer field (seconds? milliseonds?) on the Offer model.
- The server would need an API to increment/set the view time for a particular "view". We should probably cap any particular view at a maximum of 30s.
- Our reporting interface could show aggregated view time for advertisers and "view time per view" or some other similar metric to allow comparing across advertisers.
- For publishers, we should show the average view time in the reporting and overview screens. This would allow comparing publishers.

**Changes to the client**

- The client will need to start a timer once the ad is in the viewport. Whenever the ad leaves the viewport, the timer stops.
- The client will need to send one or more updates about how much an ad has been viewed. Ideally, this would be `onunload` but it could send it "periodically" (exponential backoff up to a max of maybe 30s) if that isn't reliable. We want to be very careful of performance here. Ads have historically been bad for web performance and we don't want to make it worse.

PR #418, opened 2021-07-20
Server changes needed for view time

davidfischer:
This adds the server changes needed to account for view time sent from the client in https://github.com/readthedocs/ethical-ad-client/pull/76. 

**Note:** these changes include a migration that adds a column to the Offer table. This should be run with extreme care in prod.

Fixes https://github.com/readthedocs/ethical-ad-server/issues/359 from a data capture perspective but the data still needs to be put into reports and such. We should probably collect data for a while before doing that.


adserver/models.py
```diff
@@ -1599,6 +1612,8 @@ class Offer(AdBase):
 
     """Contains data on ad views."""
 
+    MAX_VIEW_TIME = 5 * 60  # seconds
```
ericholscher, line 1615: The client also supports this, but we're validating on the server as well in case of misbehaving clients?

adserver/views.py
```diff
@@ -641,6 +641,12 @@ def get(self, request, advertisement_id, nonce):
             advertisement.track_impression(
                 request, self.impression_type, publisher=publisher, offer=offer
             )
+        elif self.impression_type == VIEWS and "view_time" in request.GET:
+            try:
+                view_time = int(request.GET["view_time"])
+                advertisement.track_view_time(offer, view_time)
+            except ValueError:
+                log.info("Invalid view time %s", request.GET["view_time"])
```
ericholscher, line 649:
```suggestion
                log.info("Invalid view time. view_time=%s", request.GET["view_time"])
```

adserver/views.py
```diff
@@ -641,6 +641,12 @@ def get(self, request, advertisement_id, nonce):
             advertisement.track_impression(
                 request, self.impression_type, publisher=publisher, offer=offer
             )
+        elif self.impression_type == VIEWS and "view_time" in request.GET:
```
ericholscher, line 644: Should we `return` from this function? It seems like we're calling all the logic here again, which feels fraught (eg. sending to analytics, etc.) 

I worry in general about the implementation of hitting the exact same view, but moving this logic higher up in the stack and returning from it would probably reduce my worry.

ericholscher (approved):
Simple enough change. The migration definitely scares me, though maybe it won't be blocking, since it's nullable?

adserver/models.py
davidfischer: This is correct. The client could send anything.

adserver/views.py
davidfischer: Views never send anything to analytics but I think your worry is a sound one. I'll see if I can make some changes to improve that.

ericholscher (approved):
Changes look good 👍

adserver/views.py
```diff
@@ -722,6 +734,35 @@ def get_response(self, request, advertisement, publisher):
         return HttpResponseRedirect(url)
 
 
+class AdViewTimeProxyView(AdViewProxyView):
```
ericholscher, line 737: What are we using from the `AdViewProxyView` that requires subclassing? Seems we're removing most of the logic?

davidfischer: Mostly just getting the relevant offer, publisher, and ad.

davidfischer:
> Sidenote: are we currently using the send_to_analytics function? I don't think we are -- might make sense to remove it.

We were at one point but I don't believe we are currently.

PR #419, opened 2021-07-22
Add permissions to see staff-only report fields

davidfischer:
- Uses Django's permission system


ericholscher (approved):
Simple enough 👍  We could even just turn these on & off for our existing users while doing the reports, which is generally useful.

davidfischer:
> Simple enough +1 We could even just turn these on & off for our existing users while doing the reports, which is generally useful.

Except that super users are considered to have all permissions. You'll have to remove superuser access.

PR #421, opened 2021-07-22
Add option of `created` sort on Staff publisher report

ericholscher:
This will make it easier to keep track of newer publishers.
I also sorted the dashboard lists by `created`,
since that's likely what is more interesting there too.


davidfischer:
So the plan is to just put the newest publishers first? I'm OK with that. My only worry is that it'll be hard to scan to find a specific publisher but it's probably fine.

davidfischer (approved):

PR #422, opened 2021-07-23
Improve difference notifications

davidfischer:
- The percent difference wasn't correct
- Set a minimum number of views (1k) before notifying

### DIFF
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -450,12 +450,15 @@ def notify_of_completed_flights():
 
 
 @app.task()
-def notify_of_publisher_changes():
-    """Send a notification when a publisher's main metrics change week to week."""
+def notify_of_publisher_changes(difference_threshold=0.25, min_views=1000):
+    """
+    Send a notification when a publisher's main metrics change week to week.
+
+    :param difference_threshold: Notify of differences larger than this (0.25 = 25%)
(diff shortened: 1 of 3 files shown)


adserver/tasks.py
```diff
@@ -478,12 +481,15 @@ def notify_of_publisher_changes():
...
+            total_views = (
+                last_week_report.total["views"] + previous_week_report.total["views"]
+            )
             last_week_value = last_week_report.total[metric]
             previous_week_value = previous_week_report.total[metric]
             if last_week_value > 0 and previous_week_value > 0:
                 metric_diff = abs((last_week_value / previous_week_value) - 1)
-                perc_diff = (last_week_value / previous_week_value) * 100
-                if metric_diff > difference_threshold:
+                perc_diff = ((last_week_value / previous_week_value) - 1) * 100
```
ericholscher, line 491: I imagine there must be a utility function to calculate this in the stdlib. Would be useful to add one if not, since we do this a few places and likely mess it up in other places :)

ericholscher (approved):

adserver/tasks.py
davidfischer: I checked `math` and `statistics` but I don't see anything.

davidfischer: I can put it into our own utilities.

PR #425, opened 2021-07-28
Interface to create a new flight

davidfischer:
This interface basically just lets you set the name and then sends you to the more complex update screen.

### DIFF
--- a/adserver/forms.py
+++ b/adserver/forms.py
@@ -354,7 +354,7 @@ class FlightCreateForm(forms.ModelForm):
 
         while Flight.objects.filter(slug=slug).exists():
             random_char = get_random_string(1)
-            slug = slugify(f"{slug}{random_char}")
+            slug = slugify(f"{slug}-{random_char}")
 
         return slug
 
--- a/adserver/forms.py
+++ b/adserver/forms.py
@@ -28,6 +28,7 @@ from django.utils.translation import ugettext_lazy as _
 from simple_history.utils import update_change_reason
 
 from .models import Advertisement
+from .models import Campaign
 from .models import Flight
 from .models import Publisher
 from .regiontopics import africa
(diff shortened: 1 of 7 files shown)


adserver/forms.py
```diff
@@ -313,6 +314,64 @@ class Meta:
...
+
+    def generate_slug(self):
+        campaign_slug = self.instance.campaign.slug
+        slug = slugify(self.instance.name)
+        if not slug.startswith(campaign_slug):
+            slug = slugify(f"{campaign_slug}-{slug}")
+
+        while Flight.objects.filter(slug=slug).exists():
+            random_char = get_random_string(1)
+            slug = slugify(f"{slug}{random_char}")
```
ericholscher, line 357:
```suggestion
            slug = slugify(f"{slug}-{random_char}")
```

To make it a bit cleaner?

adserver/forms.py
```diff
@@ -313,6 +314,64 @@ class Meta:
...
+            Fieldset(
+                "",
+                "name",
+                "campaign",
+                css_class="my-3",
+            ),
+            Submit("submit", _("Create flight")),
+            HTML(
+                "<p class='form-text small text-muted'>"
+                + str(_("You will be able to edit details on the next screen"))
```
ericholscher, line 344: 👍

adserver/templates/adserver/advertiser/flight-list.html
```diff
@@ -18,6 +18,16 @@
 <h1>{% block heading %}{% trans "Manage advertising flights" %}{% endblock heading %}</h1>
 
 
+<aside class="col text-right">
+  {% if request.user.is_staff %}
+    <a href="{% url 'flight_create' advertiser.slug %}" class="btn btn-sm btn-outline-secondary mb-3" role="button" aria-pressed="true">
+      <span class="fa fa-plus mr-1" aria-hidden="true"></span>
+      <span>{% trans 'Create flight' %}</span>
```
ericholscher, line 25: Should the UI make it more obvious this is an admin-only feature currently? Maybe a specific color or lock icon or something?

adserver/templates/adserver/advertiser/flight-list.html
```diff
@@ -18,6 +18,16 @@
 <h1>{% block heading %}{% trans "Manage advertising flights" %}{% endblock heading %}</h1>
 
 
+<aside class="col text-right">
+  {% if request.user.is_staff %}
```
ericholscher, line 22: Should this use the video recording permission, so users don't get confused and want it?

ericholscher (approved):
Simple enough. I do think we should probably do a bit more to differentiate staff actions in the UI, since we've been confused here before (eg. on recording videos -- we should probably use a similar permission so users don't ask questions).

adserver/templates/adserver/advertiser/flight-list.html
davidfischer: Probably just check "flight create" permissions more precisely.

adserver/templates/adserver/advertiser/flight-list.html
davidfischer: Longer term, I do think we'll be opening these up to users but maybe an indication isn't bad. I'll see what I can do.

davidfischer: I added a few indicators -- both color and icon.

PR #427, opened 2021-08-03
Update GeoIP Links

decaffeinatedio:
Running `make geoip` generates an error related to gzip file encoding because the files from last year are no longer live. 

Updating to this month as a quick fix, but there might be a better way to automatically populate the most recent YYYY-MM.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,8 @@
 GEOIP_DIR = geoip
 GEOIP_CITY_FILE = dbip-city-lite.mmdb.gz
 GEOIP_COUNTRY_FILE = dbip-country-lite.mmdb.gz
-GEOIP_CITY_DB_URL = https://download.db-ip.com/free/dbip-city-lite-2020-07.mmdb.gz
-GEOIP_COUNTRY_DB_URL = https://download.db-ip.com/free/dbip-country-lite-2020-07.mmdb.gz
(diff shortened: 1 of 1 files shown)


davidfischer:
Thanks. This seems fine. In our case, we don't actually use the databases from db-ip.com but this is useful to anyone developing this.

Out of curiosity, what are you using the ad server for?

decaffeinatedio:
@davidfischer Thanks!

> Out of curiosity, what are you using the ad server for?

Nothing yet, but exploring it as an alternative to Google DFP/GAM for direct deals.

ericholscher:
@decaffeinatedio Sounds cool :) Of note, we're considering offering hosted EA as a SaaS-type service, if that would make things easier for you. We'd love to have more folks contributing to the code of course, but also something we've been scheming to reduce the overhead. Definitely not too far along, but let us know if you want to chat more about this via a call or email. 👍

decaffeinatedio:
@ericholscher Ah, that's along the lines of what I was trying to experiment with; only at the sketch-of-a-landing-page stage. Happy to share thoughts, though I may not be a good prospect/customer!  

ian (at) decaffeinated.io

decaffeinatedio:
@ericholscher Would you mind pinging me your work email? Just wanted to share something I've been working on.

Issue #426, opened 2021-08-03
Error when running `make dockerprod`


decaffeinatedio:
Hi, 

Cool project! I've been able to get a local version up and running (on a pretty vanilla 21.04 box) by executing a few commands in sequence:

- `make dockerbuild`
- `make dockershell` then running a migration and superuser creation
- `make dockerserve`

Now I want to see if I can get it running in a cloud environment. However, I am running into several (stubborn) build issues with `make dockerprod`. 

First, I encountered as issue related to Alpine and the cryptography package, which I could resolve by adding `ENV CRYPTOGRAPHY_DONT_BUILD_RUST=1` into `docker-compose/production/django/Dockerfile`. 

But now I'm running into an error during the the `RUN npm install` step that I can't seem to overcome: 

```
npm ERR! code 1
npm ERR! git dep preparation failed
npm ERR! command /usr/bin/node /usr/lib/node_modules/npm/bin/npm-cli.js install --force --cache=/root/.npm --prefer-offline=false --prefer-online=false --offline=false --no-progress --no-save --no-audit
npm ERR! npm WARN using --force Recommended protections disabled.
npm ERR! npm ERR! code 1
npm ERR! npm ERR! path /root/.npm/_cacache/tmp/git-clone-3208764b/node_modules/node-sass
npm ERR! npm ERR! command failed
npm ERR! npm ERR! command sh -c node scripts/build.js
npm ERR! npm ERR! Building: /usr/bin/node /root/.npm/_cacache/tmp/git-clone-3208764b/node_modules/node-gyp/bin/node-gyp.js rebuild --verbose --libsass_ext= --libsass_cflags= --libsass_ldflags= --libsass_library=
npm ERR! npm ERR! gyp info it worked if it ends with ok
npm ERR! npm ERR! gyp verb cli [
npm ERR! npm ERR! gyp verb cli   '/usr/bin/node',
npm ERR! npm ERR! gyp verb cli   '/root/.npm/_cacache/tmp/git-clone-3208764b/node_modules/node-gyp/bin/node-gyp.js',
npm ERR! npm ERR! gyp verb cli   'rebuild',
npm ERR! npm ERR! gyp verb cli   '--verbose',
npm ERR! npm ERR! gyp verb cli   '--libsass_ext=',
npm ERR! npm ERR! gyp verb cli   '--libsass_cflags=',
npm ERR! npm ERR! gyp verb cli   '--libsass_ldflags=',
[...]

PR #428, opened 2021-08-03
Note that the prod dockerfile is unmaintained

davidfischer:
Fixes https://github.com/readthedocs/ethical-ad-server/issues/426


ericholscher (approved):
This should probably update the installation docs where we mention it as well.

PR #431, opened 2021-08-05
Add StaffRegionReport

ericholscher:
This will be exactly like the StaffGeoReport,
except have ~90% less data which should make it much faster.

It also will allow us to easily see things by region,
instead of having to guess at aggregations of countries.


davidfischer:
The tests failing are because our coverage dropped for some reason. I'm not sure what caused it but it should be fixed up in some later commits. Let me know if this is ready for a look.

ericholscher:
@davidfischer rebased this, and it should be ready for review to move forward with it 👍

ericholscher:
> One question, is whether we should add publisher to this Region index, and use it to replace the Country index after testing, or should we ship it with just the ad & region, which will make it really fast and useful for a staff report.

To answer my own question -- I think we do:

* Add publisher to the Region reports, to replace the Country index
* Add advertiser to the RegionTopic index, and use that for all our staff reporting (regions and/or topics w/ eCPM)
* Improve filtering on both of these to make them more useful

adserver/mixins.py
```diff
@@ -236,15 +244,16 @@ def get_context_data(self, **kwargs):  # pylint: disable=missing-docstring
...
         # TODO: Backport similar logic to the base report class?
         kwargs = {}
-        for arg in ["keyword", "country", "publisher"]:
-            if arg in self.request.GET:
+        for arg in ["keyword", "country", "publisher", "region", "topic"]:
+            if arg in self.request.GET and self.request.GET[arg]:
                 kwargs[arg] = self.request.GET[arg]
-                filtered = arg
+                filtered.append(arg)
+        log.info(f"Filtering report by {kwargs}")
```
davidfischer, line 256:
```suggestion
        log.debug(f"Filtering report by {kwargs}")
```

adserver/mixins.py
```diff
@@ -255,7 +264,7 @@ def get_context_data(self, **kwargs):  # pylint: disable=missing-docstring
 
         # Sort by date when filtering a specific value,
         # otherwise handle sorting via the users input
-        if filtered:
+        if len(filtered) >= self.FILTER_COUNT:
```
davidfischer, line 267: I don't really understand this. What is this threshold for? I understand we almost always want to order by date but I don't understand this.

adserver/router.py
```diff
@@ -0,0 +1,41 @@
...
+from .models import KeywordImpression
+from .models import PlacementImpression
+from .models import RegionImpression
+from .models import RegionTopicImpression
+from .models import UpliftImpression
+
+
+class IndexRouter:
+    """
+    A database router that allows for indexing into a different postgres DB than core operations.
```
davidfischer, line 12: Is the plan that this would somehow be optionally enabled somehow?

adserver/tasks.py
```diff
@@ -83,32 +84,83 @@ def _default_filters(impression_type, start_date, end_date):
 
 
 @app.task()
-def daily_update_geos(day=None):
+def daily_update_geos(day=None, geo=True, region=True):
```
davidfischer, line 87: I realize there's some common code, but combining this feels awkward and hard to test. I wouldn't mind just seeing two tasks.

adserver/templates/adserver/reports/includes/publisher-report-table.html
```diff
@@ -12,7 +12,7 @@
       {% if "adserver.staff_publisher_fields" in perms %}
         <th class="text-right staff-only"><strong>{% trans 'Our&nbsp;Rev' %}</strong></th>
         <th class="text-right staff-only"><strong>{% trans 'Total&nbsp;Rev' %}</strong></th>
-        <th class="text-right staff-only"><strong>{% blocktrans %}<abbr title="Effective cost per thousand impressions">eCPM</abbr>{% endblocktrans %}</strong></th>
+        <th class="text-right staff-only"><strong>{% blocktrans %}<abbr title="Effective cost per thousand impressions">Total eCPM</abbr>{% endblocktrans %}</strong></th>
```
davidfischer, line 15:
```suggestion
        <th class="text-right staff-only"><strong>{% blocktrans %}Total <abbr title="Effective cost per thousand impressions">eCPM</abbr>{% endblocktrans %}</strong></th>
```

davidfischer:
The test errors look real this time.

adserver/mixins.py
ericholscher: Basically there is a report with 2 indexes, so it makes stuff weird :/

adserver/router.py
ericholscher: Woops. I had it locally to try and generate reports against prod data but locally (so I could do migrations), but hit issues with the FK. I wonder about denormalizing all the FK data on these reports -- would make them faster, and basically the only thing we're pulling from the advertiser relation is the pricing (CPC/CPM and cost) as well as the campaign type (house/community/paid). Might be useful to just store that on the record.

adserver/tasks.py
ericholscher: Pretty soon the geos part will go away, I think. I just didn't want to run the exact same query twice, but I can split it up if we feel strongly. I think the real path forward is deleting the old one after we're happy w/ the new reporting.

ericholscher:
@davidfischer The tests are passing, just broken on coverage. Ready for another review.

tox.ini
```diff
@@ -42,4 +42,4 @@ whitelist_externals = echo
 commands =
     # Run html first so we have a report of the failing coverage results
     coverage html
-    coverage report --show-missing --fail-under=95
+    coverage report --show-missing --fail-under=93
```
ericholscher, line 45: I appreciate this check, but feels like this code is as tested as I'd want. I can add a bunch more small ones, but I think this is probably more correct. I don't feel super strongly tho.

I can probably also fix it by deleting the `rtdimport` script and some others we aren't using.

adserver/tasks.py
davidfischer: I agree with deleting the geo report stuff afterwards but I feel like it would be cleaner then to have a separate function that we just blow away rather than refactoring this to work 2 ways and then un-factoring it.

tox.ini
davidfischer: That's fine. We have some code to delete anyway. My goal with setting the minimum at 95 was to make sure we didn't slide back in coverage after getting to 95%.

davidfischer (approved):

adserver/tasks.py
ericholscher: Yea, definitely cleaner, but uses more DB resources. I was opting to not hammer the DB as hard, but it's definitely a tradeoff.

PR #433, opened 2021-08-11
Explicitly addresses the SaaS approach

MxRaCohen:
As we define/refine our ability to support 3rd party ad/sponsorship administration, we wanted to let everyone know that this is on our roadmap.


davidfischer:
This error seems to be transient because we're "very" close to being below our testing threshold. This should be resolved by some of the existing PRs or if not I'll take a look.

davidfischer (approved):

PR #435, opened 2021-08-12
Remove server side analytics which we weren't using

davidfischer:


ericholscher (approved):
I think this is useful, since folks might see this and think we're using GA on the backend. I don't think we'll do that in the future, so 👍  on this change.

PR #441, opened 2021-08-18
Tweaks payouts with issues that we've found

ericholscher:
This includes:

* Ability to set a publisher to have skipped payouts
* Better filtering on the payout dashboard
* Removes some of the less user features in publisher output


adserver/staff/views.py
```diff
@@ -81,54 +81,75 @@ class PublisherPayoutView(StaffUserMixin, TemplateView):
...
+    CACHE_SECONDS = 3600 * 24
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
-        limit = int(self.request.GET.get("limit", 0))
-        all_publishers = self.request.GET.get("all")
         publisher_slug = self.request.GET.get("publisher")
+        first = self.request.GET.get("first", "")
+        paid = self.request.GET.get("paid", "")
```
davidfischer, line 98:
```suggestion
        paid = self.request.GET.get("paid", "") == "paid"
```

You might consider something like this to just make it a boolean.

adserver/templates/adserver/reports/includes/publisher-report-table.html
```diff
@@ -12,7 +12,7 @@
       {% if "adserver.staff_publisher_fields" in perms %}
         <th class="text-right staff-only"><strong>{% trans 'Our&nbsp;Rev' %}</strong></th>
         <th class="text-right staff-only"><strong>{% trans 'Total&nbsp;Rev' %}</strong></th>
-        <th class="text-right staff-only"><strong>{% blocktrans %}<abbr title="Effective cost per thousand impressions">eCPM</abbr>{% endblocktrans %}</strong></th>
+        <th class="text-right staff-only"><strong>{% blocktrans %}<abbr title="Effective cost per thousand impressions">Total eCPM</abbr>{% endblocktrans %}</strong></th>
```
davidfischer, line 15:
```suggestion
        <th class="text-right staff-only"><strong>{% blocktrans %}Total <abbr title="Effective cost per thousand impressions">eCPM</abbr>{% endblocktrans %}</strong></th>
```

davidfischer (approved):
Looks reasonable. I made a few suggestions but otherwise seems good.

adserver/staff/views.py
ericholscher: I want 3 states, which I guess could be True/False/None.

PR #444, opened 2021-08-20
More tweaks to publisher notifications

davidfischer:
Increases the threshold of number of views to trigger a notification. Based on this, we would have gotten just a handful of notifications in the last week.


ericholscher (approved):
Seems like we also want to take into account newness, but we can do that later. I'd like to find a way to report on publishers outside of normal bounds (eg. .05-.5 CPM), so we can keep track of it -- but it definitely requires some more thinking.

PR #454, opened 2021-09-20
Add ability to export region data

ericholscher:
This was to let me import it into Google Docs,
where I could do some test graphing on it.
I didn't get a ton of data out of it,
but it at least helped me understand what is happening.

### DIFF
--- a/adserver/reports.py
+++ b/adserver/reports.py
@@ -40,7 +40,9 @@ class BaseReport:
     # Fields to apply select_related on the queryset
     select_related_fields = ("advertisement", "advertisement__flight")
 
-    def __init__(self, queryset, index=None, order=None, max_results=None, **kwargs):
+    def __init__(
+        self, queryset, index=None, order=None, max_results=None, export=False, **kwargs
+    ):
         """
         Initialize the report using filtered, ordered queryset.
 
@@ -49,6 +51,7 @@ class BaseReport:
             This is sometimes necessary for example when filtering a geo report to a specific country,
             you probably want to index by date.
         :param order: Override the report result order.
+        :param export: Tell the report if it's being exported, so it can adjust display for processing
         :param max_results: Override the maximum results to return
         """
         self.queryset = queryset
@@ -61,6 +64,8 @@ class BaseReport:
             self.max_results = max_results
         else:
             self.max_results = self.DEFAULT_MAX_RESULTS
+
+        self.export = export
 
         # Save any other keyword args for use in subclasses
         self.kwargs = kwargs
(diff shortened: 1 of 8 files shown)


adserver/templates/adserver/dashboard.html
```diff
@@ -20,13 +20,13 @@ <h2>{% trans 'Staff Reports' %}</h2>
...
+        <li><a href="{% url 'staff_regiontopic_report' %}">{% trans 'Staff Region & Topic Report' %}</a> (<a href="{% url 'staff_regiontopic_report' %}?start_date={{ month_start|date:'Y-m-d' }}">{% trans 'This month' %}</a>)</li>
+        <li><a href="{% url 'staff_keyword_report' %}">{% trans 'Staff Keyword Report' %}</a> (<a href="{% url 'staff_keyword_report' %}?start_date={{ month_start|date:'Y-m-d' }}">{% trans 'This month' %}</a>)</li>
+        <li><a href="{% url 'staff_geo_report' %}">{% trans 'Staff Geo Report' %}</a> (<a href="{% url 'staff_geo_report' %}?start_date={{ month_start|date:'Y-m-d' }}">{% trans 'This month' %}</a>)</li>
+        <li><a href="{% url 'staff_region_report' %}">{% trans 'Staff Region Report' %}</a> (<a href="{% url 'staff_region_report' %}?start_date={{ month_start|date:'Y-m-d' }}">{% trans 'This month' %}</a>)</li>
+        <li><a href="{% url 'publisher_uplift_report' %}">{% trans 'Staff Uplift Report' %}</a> (<a href="{% url 'publisher_uplift_report' %}?start_date={{ month_start|date:'Y-m-d' }}">{% trans 'This month' %}</a>)</li>
```
ericholscher, line 29: This adds a `This month` view to all the staff reports. I kept hitting issues where they were filtering by a huge date range from a previous view.

adserver/reports.py
```diff
@@ -288,7 +293,11 @@ def get_index_display(self, index):
         """Handle making sure dates use the same 3 letter syntax as Django."""
         if isinstance(index, datetime.date):
             # %a is the day of week, so you know which are weekends
-            return index.strftime("%b %d, %Y (%a)")
+            value = index.strftime("%b %d, %Y")
+            # Make it so these parse as dates in a CSV
+            if not self.export:
+                value += index.strftime(" (%a)")
```
davidfischer, line 299: If we're already going to differentiate, I'd just export as ISO8601 and then keep the above format for display. If we want to get really fancy, we could use an [HTML `time`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time)

adserver/views.py
```diff
@@ -1970,7 +1986,18 @@ class StaffRegionReportView(AllReportMixin, BaseReportView):
 
     """An region report for all publishers."""
 
-    fieldnames = ["index", "views", "clicks", "ctr"]
+    export_view = "staff_region_report_export"
```
davidfischer, line 1989: This is causing the test failure although that check should just be removed. The check was ensuring that "CSV Export" didn't appear since it wasn't implemented.

davidfischer (approved):

adserver/reports.py
ericholscher: Yea, something to think more about. Don't think we need to address it here.

PR #457, opened 2021-09-20
Add a read replica DB router & settings

ericholscher:
This will allow us to use a read replica in production.
Currently this is only used for reports,
and in the task that cleans Offers from the DB.
In the future though,
we will extend the usage to most tasks and non-write operations.

The biggest issue here is making sure we don't have replica lag.
Things like `get_or_create` are thus using the default DB explicitly,
when acting on Impression models.

### DIFF
--- a/adserver/management/commands/archive_offers.py
+++ b/adserver/management/commands/archive_offers.py
@@ -84,11 +84,6 @@ class Command(BaseCommand):
         zipped_output_file = Path(str(output_file) + ".bz2")
         end_day = day + datetime.timedelta(days=1)
 
-        # Find the proper DB to query
-        copy_connection = "default"
-        if "replica" in connection:
-            copy_connection = "replica"
-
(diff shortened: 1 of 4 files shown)


ericholscher:
Not sure how to actually fix this pylint error :(

davidfischer (approved):
This looks good. There's a bit of complication here to get around replica lag as you noted. However, I suspect (but couldn't easily confirm) that the `get_or_create` logic should *just work* at least as far as daily tasks are concerned. Is that not so?

davidfischer:
I also fixed up the linting

ericholscher:
> I also fixed up the linting

Weird. I tried it on the function itself https://github.com/readthedocs/ethical-ad-server/pull/457/commits/9bdab9f975cd154eac122dec3c60496a86091452 -- but that didn't seem to work either. I don't fully understand pylint, heh.

PR #458, opened 2021-09-22
Fixes a bug with old-style ads

davidfischer:
By submitting an ad with no content, it was possible to convert your ad to an old-style ad (a single HTML text field). This fixes that by making the content field required for new ads and the text field required for old ads.


ericholscher (approved):
Simple enough 👍

PR #462, opened 2021-09-29
Notify when daily reports are aggregated

davidfischer:
Sends a Slack when the daily reports are aggregated successfully.

### DIFF
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -463,6 +463,15 @@ def update_previous_day_reports(day=None):
     daily_update_uplift(start_date)
     daily_update_regiontopic(start_date)
 
+    # Send notification to Slack about previous day's reports
+    if settings.SLACK_TOKEN:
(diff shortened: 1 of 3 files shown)


adserver/tasks.py
```diff
@@ -463,6 +463,15 @@ def update_previous_day_reports(day=None):
     daily_update_uplift(start_date)
     daily_update_regiontopic(start_date)
 
+    # Send notification to Slack about previous day's reports
+    if settings.SLACK_TOKEN:
```
ericholscher, line 467: Might be worth moving this the `slack_message` code, so we don't need it everywhere?

ericholscher (approved):
Simple and effective! 👍

adserver/tasks.py
davidfischer: Cool. It's a third party package but I'll move a wrapper into utils.

PR #465, opened 2021-09-30
Archive offers daily task

davidfischer:
Creates a task which archives a single day's (90d ago) offers every day.


adserver/management/commands/archive_offers.py
```diff
@@ -93,10 +94,11 @@ def handle_archive_day(self, day):
                 WHERE date >= '{day:%Y-%m-%d}' AND date < '{end_day:%Y-%m-%d}'
                 ORDER BY date
             ) TO STDOUT WITH CSV HEADER"""
-        with connection[settings.REPLICA_SLUG].cursor() as cursor:
+        with connections[settings.REPLICA_SLUG].cursor() as cursor:
```
davidfischer, line 97: I'm not 100% sure how this was working before...

adserver/tasks.py
```diff
@@ -588,3 +590,38 @@ def notify_of_publisher_changes(difference_threshold=0.25, min_views=10_000):
                             ),
                         },
                     )
+
+
+@app.task(time_limit=60 * 60 * 4)
+def archive_offers(day=None, delete=False):
```
davidfischer, line 596: I figure switch this to `delete=True` after we deploy this and see how it goes for a few days.

adserver/management/commands/archive_offers.py
ericholscher: I patched it locally heh

adserver/management/commands/archive_offers.py
```diff
@@ -181,13 +183,13 @@ def delete_offers(self, day):
...
         # Always delete from the default.
-        with connection["default"].cursor() as cursor:
+        with connections["default"].cursor() as cursor:
             # Offers are normally immutable so the delete has to be run as raw sql
             cursor.execute(
                 query,
                 [day, end_day],
             )
-            deleted_offers = cursor.fetchone()
+            deleted_offers = cursor.fetchone() or 0
```
ericholscher, line 192: I also hit this one :)

adserver/tasks.py
```diff
@@ -588,3 +590,38 @@ def notify_of_publisher_changes(difference_threshold=0.25, min_views=10_000):
...
+    start_date, _ = _get_day(day)
+
+    args = [
+        "-s",
+        start_date.strftime("%Y-%m-%d"),
+        "-e",
+        start_date.strftime("%Y-%m-%d"),
+    ]
+    if delete:
+        args += ["-d"]
```
ericholscher, line 615: I like using long-form args in scripts, so we can easily tell what it's doing:

```suggestion
        args += ["--delete-offers"]
```

config/settings/production.py
```diff
@@ -134,6 +134,10 @@
         "task": "adserver.tasks.notify_of_completed_flights",
         "schedule": crontab(hour="0", minute="0"),
     },
+    "every-day-archive-offers": {
+        "task": "adserver.tasks.archive_offers",
+        "schedule": crontab(hour="4", minute="0"),
```
ericholscher, line 139: I worry a little bit about this overlapping with the indexing tasks. It might be nice to implement a locking pattern here at some point, so we don't overload the DB.

ericholscher (approved):

config/settings/production.py
davidfischer: That's fair. I think the slack messages will help us see if that's happening.

adserver/management/commands/archive_offers.py
mattishaden: Shouldn't this be:

> deleted_offers = cursor.rowcount

fetchone() for a DELETE query returns an exception:

> django.db.utils.ProgrammingError: no results to fetch

See discussion in: https://dba.stackexchange.com/questions/75921/delete-returns-nothing-in-psycopg2

davidfischer: @mattishaden thanks for catching this. You are correct. I'm not sure how this every worked although we did run it multiple times. Curious...

davidfischer:
We are holding off on this. We are going to close it and re-work if it is needed.

mattishaden:
So this is not running be default in latest version? What path did you take for the offers table? Looking for some pointer on how we should handle the table size, which is getting out of hand:

```
 schemaname |                   tablename                    |  Size  
------------+------------------------------------------------+---------
 public     | adserver_offer                                 | 140 GB
 public     | adserver_click                                 | 88 MB
 public     | adserver_adimpression                          | 40 MB
```

What would happen if we just truncated it?

ericholscher:
@mattishaden We are creating offers tables quarterly and then `pg_dump`'ing the old ones as a backup: https://github.com/readthedocs/ethical-ad-server/blob/d6ff4a03b839aaf1cd292ce868e85c2200c22df2/config/settings/base.py#L130-L151

PR #474, opened 2021-10-20
Use impression-caching on RegionTopic report to improve performance

ericholscher:
This is a proof of concept attempt at denormalizing the data for the report onto the impression.
This makes the DB query much lighter,
and doesn't have any joins.

It also is neat because it lets me index production ad data into a local DB, since there aren't any ForeignKey's in the Impression data.

I think we could migrate most of our staff reports to this format,
which would have a few benefits:

* Smaller tables since all flights with the same pricing are aggregated
* Faster queries because there are no joins

I think this could give us a lot of benefits for BI-style uses.

We'd still need publisher & advertiser FK's on the reports that we put on the dashboards,
but we could even split these up to make them faster..

### DIFF
--- a/adserver/models.py
+++ b/adserver/models.py
@@ -1603,16 +1603,16 @@ class RegionTopicImpression(BaseImpression):
 
     region = models.CharField(_("Region"), max_length=100)
     topic = models.CharField(_("Topic"), max_length=100)
-    advertisement = models.ForeignKey(
-        Advertisement,
-        related_name="regiontopic_impressions",
-        on_delete=models.PROTECT,
-        null=True,
+    cpc = models.DecimalField(
+        _("Cost Per Click"), max_digits=5, decimal_places=2, default=0
+    )
+    cpm = models.DecimalField(
+        _("Cost Per 1k Impressions"), max_digits=5, decimal_places=2, default=0
     )
 
     class Meta:
         ordering = ("-date",)
-        unique_together = ("date", "region", "topic", "advertisement")
+        unique_together = ("date", "region", "topic", "cpc", "cpm")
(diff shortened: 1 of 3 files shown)


davidfischer:
> > It would make things a lot faster (eg. ~30 entries per day per advertiser)
> 
> This is definitely not the case. We'd have a record for each day for each publisher/advertiser combo. So it would be targeted publishers * advertisers * region * topic.

I think you misunderstood me. This is only the case if we make separate tables for advertisers and publishers. In that case, there would be ~30 records per advertiser in the advertiser table and ~30 records per publisher in the other table. However, it doesn't seem that's what you were proposing.

ericholscher:
> I think you misunderstood me. This is only the case if we make separate tables for advertisers and publishers. In that case, there would be ~30 records per advertiser in the advertiser table and ~30 records per publisher in the other table. However, it doesn't seem that's what you were proposing.

I was mostly proposing staff-level indexes and user-facing indexes. I don't think we want 3 total, but it would be useful if we could aggregate the staff reports without the publisher or advertiser, to make them super fast. 

Then for user-facing reports, since they are always filtered by a single publisher or advertiser, I'm less worried about the size of the indexes. In that case, I think that we already have is fine, and seems mostly fast enough. The AdImpression table for example seems to work fine with years of data in it, on the Revenue reports.

ericholscher:
I'm going to close this PR though. I don't think this is the problem we need to be discussing in depth currently. I just thought it was a neat example of how easy it would be to denormalize the data on the Impression. We should focus on Offers, and shipping the RegionImpression's and TopicImpressions to users.

PR #475, opened 2021-10-21
Charting/graphing with metabase

davidfischer:
Metabase is a graphing/BI tool. In this PR, this just adds 2 graphs on the Publisher Overview screen and the Advertiser Overview screen. These are currently restricted to staff but if rollout goes well, we'll open them up.

- Metabase can be used locally in development with Docker. Production will probably be metabase cloud.
- Data can be explored by technical and non-technical staff. Queries can be saved internally or published for embedding in the application.
- It allows embedding graphs that can have their parameters (eg. advertiser/publisher/start-date) locked down. Changing parameters requires the signing key.
- Adds a management command that is useful in exporting/versioning graphs/queries from metabase (either prod or development).

### DIFF
--- a/adserver/templates/adserver/advertiser/overview.html
+++ b/adserver/templates/adserver/advertiser/overview.html
@@ -2,6 +2,7 @@
 {% load crispy_forms_tags %}
 {% load humanize %}
 {% load i18n %}
+{% load metabase %}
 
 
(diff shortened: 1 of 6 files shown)


davidfischer:
> Don't love the load/dump process, but seems reasonable enough.

I actually don't think we'll use it much. I more see it as a way to version control these things.

PR #483, opened 2021-11-08
Enable Stripe billing portal for advertisers

davidfischer:
- They can't do anything other than view past invoices

### DIFF
--- a/adserver/templates/adserver/base.html
+++ b/adserver/templates/adserver/base.html
@@ -108,6 +108,16 @@
                   <span>{% trans 'Authorized users' %}</span>
                 </a>
               </li>
+
+              {% if advertiser.stripe_customer_id %}
+              <li class="nav-item">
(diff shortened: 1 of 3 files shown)


ericholscher (approved):
Simple enough 👍 

I'm curious what more we can do to integrate Stripe at some point, but more on the publisher side than advertisers to start for payouts. I dislike how disconnected payouts are from the actual money moving :/

davidfischer:
If I were doing it, I'd probably start with the [django-stripe package](https://dj-stripe.readthedocs.io/en/master/). The package receives some amount of blessing from stripe. Essentially, it syncs all the stripe data into your local database and sets up a webhook to receive updates from stripe. Then you can just foreign key to "stripe" models (eg. `customer = models.ForeignKey('djstripe.Customer')`). This turns all read-only stripe operations into database operations using the usual ORM. I believe it has reasonable support for Stripe Connect which we use for publishers.

PR #484, opened 2021-11-10
Fix another silly month/year date bug

ericholscher:
This was causing publishers created 1 year ago to not show a pending balance.


davidfischer (approved):
I added to this PR so my approval shouldn't count for everything. However, if you agree with my changes @ericholscher, let's merge it!

ericholscher:
Looks good 👍

PR #492, opened 2021-12-01
Update exclude list

davidfischer:


ericholscher (commented):
Should `IN` be added to another region?

davidfischer:
I don't think so. Advertisers do want it but they say that explicitly when they do.

ericholscher:
It just feels weird to have it outside of a region list for reporting. Generally we want each of the countries to be in one of the lists of regions, otherwise we don't know where to include it in reports.

davidfischer:
I think you're correct. I'll add a separate South Asia region.

ericholscher (approved):
💯

PR #494, opened 2022-01-11
Initial DJStripe integration

davidfischer:
This initial PR only adds the stripe models to our database. 

## How does this work
[DJStripe](https://dj-stripe.readthedocs.io/) is a Stripe sponsored but independent project since Aug 2020. It works by having local database models that mirror the data in Stripe. When Stripe receives changes (say a customer pays an invoice via ACH), Stripe will send a webhook to our djstripe so our local copy stays up to date. In this way, it will be possible for our application to check whether invoices have been paid or are overdue as easily as querying the database.

## Testing locally

* Set the Stripe test key from https://dashboard.stripe.com/test/apikeys into `.envs/local/django` in the `STRIPE_SECRET_KEY` setting.
* Run the following:
    ```
    make dockerbuild      # rebuild your docker image with djstripe and deps
    make dockershell      # get a shell to the docker image
    ./manage.py migrate   # create the stripe tables in the DB
    ./manage.py djstripe_sync_models      # sync data from stripe to the DB
    ```
* Browse the Stripe data in the Django admin: http://ethicaladserver.lan:5000/admin/djstripe/

## Deploying to prod

* We already set the stripe secret key in prod so we don't need to do anything there.
* Create a Stripe webhook in https://dashboard.stripe.com/webhooks that points to `https://server.ethicalads.io/stripe/webhook/`
* Set the webhook secret into `DJSTRIPE_WEBHOOK_SECRET`

## Next steps
The next steps are:

* Create FKs from our models to these Stripe models:
    * Advertiser -> stripe.Customer
    * Flight m2m-> stripe.Invoice
    * Flight -> stripe.Coupon?
    * Publisher -> stripe.Account?
* Migrate our existing data from using stripe IDs to use the FKs

### DIFF
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -1,6 +1,5 @@
 """Celery tasks for the ad server."""
 import datetime
-import json
 import logging
 from collections import defaultdict
 
@@ -285,12 +284,9 @@ def daily_update_keywords(day=None):
             ):
                 continue
 
-            keywords = json.loads(values["keywords"])
-            publisher_keywords = set(keywords)
+            publisher_keywords = set(values["keywords"])
 
-            flight_targeting = json.loads(
-                values["advertisement__flight__targeting_parameters"]
-            )
+            flight_targeting = values["advertisement__flight__targeting_parameters"]
             flight_keywords = set(flight_targeting.get("include_keywords", {}))
 
             matched_keywords = publisher_keywords & flight_keywords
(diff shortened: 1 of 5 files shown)


davidfischer:
The test failure here is real and is related to the required jsonfield upgrade. I'm working on a fix.

ericholscher (approved):
Looks simple enough 👍

PR #495, opened 2022-01-12
Disable metabase restart

ericholscher:
This caused metabase to start whenever I started docker,
which was causing a lot of CPU overhead.

I'm curious if we even want metabase locally by default?
I've never used it, and it seems like a heavy thing to run on all dev instances.

### DIFF
--- a/docker-compose-local.yml
+++ b/docker-compose-local.yml
@@ -88,7 +88,6 @@ services:
 
   metabase:
     image: metabase/metabase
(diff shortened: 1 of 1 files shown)


davidfischer:
I was going to send this same PR. Metabase locally definitely causes some issues sometimes. If it continues, let's just disable it locally.

davidfischer (approved):

PR #498, opened 2022-01-17
Stripe foreign key fields migrations

davidfischer:
Migrate our existing storage of Stripe IDs to django-stripe FKs. Includes both a schema and data migration.


adserver/migrations/0063_data_stripe_fk_migration.py
```diff
@@ -0,0 +1,53 @@
...
+def forwards(apps, schema_editor):
+    """Migrate to proper Stripe FKs."""
+    Advertiser = apps.get_model("adserver", "Advertiser")
+    Publisher = apps.get_model("adserver", "Publisher")
+    Customer = apps.get_model("djstripe", "Customer")
+    Account = apps.get_model("djstripe", "Account")
+
+    for advertiser in Advertiser.objects.all():
+        if advertiser.stripe_customer_id:
+            advertiser.djstripe_customer = Customer.objects.get(id=advertiser.stripe_customer_id)
```
ericholscher, line 20: I wonder if we need to be defensive here at all and check for `None`? Perhaps `filter().first()`?

It seems like there could be old/outdated id's in the DB that aren't in Stripe for some reason.

adserver/models.py
```diff
@@ -157,6 +158,16 @@ class Publisher(TimeStampedModel, IndestructibleModel):
...
+    djstripe_account = models.ForeignKey(
+        djstripe_models.Account,
+        verbose_name=_("Stripe connected account"),
+        on_delete=models.SET_NULL,
+        blank=True,
+        null=True,
+        default=None,
+    )
+
+    # Deprecated - migrate to `stripe_account`
```
ericholscher, line 170: ❤️

config/settings/base.py
```diff
@@ -368,8 +365,17 @@
     # However, the app won't start without the secret
     # with this setting set to the default
     DJSTRIPE_WEBHOOK_VALIDATION = "retrieve_event"
-if STRIPE_LIVE_SECRET_KEY:
-    INSTALLED_APPS += ["djstripe"]
+
+# This is a bit convoluted but basically django-stripe raises a critical
+# if the key isn't set to anything. This prevents migrations from being created
+# among other things.
+# https://github.com/dj-stripe/dj-stripe/issues/1360
```
ericholscher, line 372: 💯 comment.

ericholscher (approved):
Solid improvement 🎉

adserver/migrations/0063_data_stripe_fk_migration.py
davidfischer: I purposefully made this *not* defensive. I want it to blow up loudly (fail the migration and force data correction) if something like this occurs.

davidfischer:
Migration 62 got created in another PR so this needs a slight tweak.

davidfischer:
Migrations were re-numbered.

PR #503, opened 2022-01-20
Upgrade to Python 3.8

davidfischer:
There was more to this than I expected but it wasn't too bad.


adserver/management/commands/rtdimport.py
```diff
@@ -296,6 +296,27 @@ def import_advertisements(self, advertisements_data):
             "small",
         ]
 
+        with open(
```
ericholscher, line 299: I assume this was automated? We can probably just delete this file.

docker-compose/local/django/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.6-alpine
+FROM python:3.8-alpine
```
ericholscher, line 1: I'd love to move this to Ubuntu to match production at some point, if it isn't hard to do.

ericholscher (approved):
Looks good :+1:

adserver/management/commands/rtdimport.py
davidfischer: I agree this file can be deleted. We're never running it again. I'll do a separate PR after merging this.

docker-compose/local/django/Dockerfile
davidfischer: It shouldn't be that hard.

stsewd (approved):

PR #505, opened 2022-01-20
Expose View Rate to publishers.

ericholscher:
I wonder if we need additional context here,
since this will almost never be 100% for various reasons (eg. a user loads a page then closes the browser).

I do think exposing this data will motivate publishers to improve their placements though,
if we can communicate it effectively.


davidfischer (approved):
This looks fine. I do think we will get some questions here and this comes back to our need to improve our onboarding. With that said, it's fine to roll out for now.

PR #508, opened 2022-01-21
Optimize the keyword aggregation

davidfischer:
Reduces the select queries from 4->1 and reduce the write queries by a factor of 4. This should significantly speed up the nightly aggregations.


ericholscher (approved):
Neat. We can probably backport these to the other reports as well if this works well.

PR #509, opened 2022-01-25
Start modeling SaaS publishers to show them billing data

ericholscher:
This enables us to expose decisions (shown as API requests) to SaaS
publishers. This is what we're currently planning to bill on, so it's the
main item that they need to understand.

Based on #505 to not have conflicts.


adserver/templates/adserver/reports/includes/publisher-report-table.html
```diff
@@ -8,13 +8,17 @@
       <th class="text-right"><strong>{% trans 'Views' %}</strong></th>
       <th class="text-right"><strong>{% trans 'Clicks' %}</strong></th>
       <th class="text-right"><strong>{% blocktrans %}<abbr title="Click through rate">CTR</abbr>{% endblocktrans %}</strong></th>
+      <th class="text-right"><strong>{% blocktrans %}<abbr title="% of traffic that viewed the ad">View Rate</abbr>{% endblocktrans %}</strong></th>
+      {% if publisher.saas %}
+      <th class="text-right"><strong>{% blocktrans %}<abbr title="Billable API requests">API Requests</abbr>{% endblocktrans %}</strong></th>
```
davidfischer, line 13: This explanation feels like it could be slightly improved but hard for me to come up with something concretely better.

davidfischer (approved):
Just so I'm clear, these will be publishers who can manage the ads that run on their platform. However, we're (for now at least) going to have a separate advertiser where they setup the ads?

adserver/templates/adserver/reports/includes/publisher-report-table.html
ericholscher: Yea... I think we need a reference doc that explains all these in more depth. Trying to put them in a tooltip is a losing battle.

ericholscher:
> Just so I'm clear, these will be publishers who can manage the ads that run on their platform. However, we're (for now at least) going to have a separate advertiser where they setup the ads?

@davidfischer Yea -- that's basically what the PSF is doing, is they get an Advertiser & Publisher account that just targets them basically. It's a bit hacky, but it works. Definitely some room for future improvements.

PR #512, opened 2022-01-27
Use Cloudflare GeoIP and IP canonicalization

davidfischer:
* Adds new middlewares:
  - one that uses [Cloudflare's IP headers](https://developers.cloudflare.com/fundamentals/get-started/http-request-headers) for IP canonicalization.
  - one that uses Cloudflare's geolocation (country only) headers for geolocation 
  - Adds a middleware that uses the GeoIP database but we probably won't use it
  - Breaks the `X-Server` header into its own middleware ~and makes it a staff-only header~
* Refactors our geolocation data to be in a dataclass attached to `request.geo`
* Adds a bunch of test cases to try to catch errors

The reasoning here is a few fold:

- Cloudflare likely has more up-to-date and complete GeoIP databases than we do
- Cloudflare will keep more up-to-date with Tor exit node lists

**NOTE**: This will be a bit challenging to fully test before it hits production. It might be worth setting up on a different domain briefly to test this change.

### DIFF
--- a/adserver/middleware.py
+++ b/adserver/middleware.py
@@ -33,6 +33,7 @@ class CloudflareMiddleware:
     Sets request.ip_address and request.country from Cloudflare headers.
 
     See: https://developers.cloudflare.com/fundamentals/get-started/http-request-headers
+    See: https://support.cloudflare.com/hc/en-us/articles/200168236-Configuring-Cloudflare-IP-Geolocation
 
     SECURITY NOTE: This middleware *SHOULD BE DISABLED* if not running on Cloudflare.
         Otherwise these headers are spoofable and could result in invalid traffic.
@@ -82,6 +83,7 @@ class CloudflareMiddleware:
         country_code = request.headers.get(self.COUNTRY_HEADER, None)
         if country_code == "XX":
             # CF always adds the header and sets "XX" if it can't geolocate
+            # CF uses "T1" for Tor traffic
             country_code = None
 
         return country_code
--- a/adserver/middleware.py
+++ b/adserver/middleware.py
@@ -5,6 +5,9 @@ import socket
 
 from django.conf import settings
 
+from .utils import GeolocationData
+from .utils import get_geoipdb_geolocation
+
 
 log = logging.getLogger(__name__)  # noqa
 
@@ -28,69 +31,35 @@ class ServerInfoMiddleware:
         return response
 
 
-class CloudflareMiddleware:
+class IpAddressMiddleware:
 
     """
-    Sets request.ip_address and request.country from Cloudflare headers.
+    Sets the IP address of the user onto the request object.
 
-    See: https://developers.cloudflare.com/fundamentals/get-started/http-request-headers
-    See: https://support.cloudflare.com/hc/en-us/articles/200168236-Configuring-Cloudflare-IP-Geolocation
-
-    SECURITY NOTE: This middleware *SHOULD BE DISABLED* if not running on Cloudflare.
-        Otherwise these headers are spoofable and could result in invalid traffic.
-        Also, Cloudflare IP Geolocation must be enabled.
+    In production, you probably want to use the CloudflareIpAddressMiddleware
+    or the XForwardedForMiddleware depending on your setup.
     """
 
-    COUNTRY_HEADER = "CF-IPCountry"
-    IP_HEADER = "CF-Connecting-IP"
+    PROVIDER_NAME = "REMOTE_ADDR header"
 
     def __init__(self, get_response):
         """One-time configuration and initialization."""
         self.get_response = get_response
 
     def __call__(self, request):
-        """Sets request.ip_address and request.country based on Cloudflare headers."""
-        request.ip_address = self._get_ip_address(request)
-        request.user_country = self._get_country(request)
+        """Sets the IP address of the user onto the request object."""
+        request.ip_address = self.get_ip_address(request)
         response = self.get_response(request)
 
-        if request.user_country and (settings.DEBUG or request.user.is_staff):
-            response["X-Adserver-GeoIPProvider"] = "Cloudflare"
-
+        if settings.DEBUG or request.user.is_staff:
+            response["X-Adserver-IpAddress-Provider"] = self.PROVIDER_NAME
         return response
 
-    def _get_ip_address(self, request):
-        ip = request.headers.get(self.IP_HEADER, None)
-
-        if ip:
-            try:
-                ipaddress.ip_address(ip)
-            except ValueError:
-                # This should be OK to log since this is a spoofed address
-                log.warning(
-                    "Invalid IP in Cloudflare headers. "
-                    "This header is probably untrustworthy so the CloudflareMiddleware should be disabled. "
-                    "%s=%s",
-                    self.IP_HEADER,
-                    ip,
-                )
-                ip = None
-        if not ip:
-            ip = request.META.get("REMOTE_ADDR", "")
-
-        return ip
-
-    def _get_country(self, request):
-        country_code = request.headers.get(self.COUNTRY_HEADER, None)
-        if country_code == "XX":
-            # CF always adds the header and sets "XX" if it can't geolocate
-            # CF uses "T1" for Tor traffic
-            country_code = None
-
-        return country_code
+    def get_ip_address(self, request):
+        return request.META.get("REMOTE_ADDR", None)
 
 
-class XForwardedForMiddleware:
+class XForwardedForMiddleware(IpAddressMiddleware):
 
     """
     Sets request.ip_address with the client's IP from x-forwarded-for.
@@ -102,20 +71,11 @@ class XForwardedForMiddleware:
         header is not guaranteed from the load balancer as a user could fake it.
     """
 
+    PROVIDER_NAME = "X-Forwarded-For"
     HEADER_NAME = "X-Forwarded-For"
 
-    def __init__(self, get_response):
-        """One-time configuration and initialization."""
-        self.get_response = get_response
-
-    def __call__(self, request):
-        """Sets request.ip_address based on x-forwarded-for."""
-        request.ip_address = self._get_ip_address(request)
-        response = self.get_response(request)
-        return response
-
-    def _get_ip_address(self, request):
-        client_ip = None
+    def get_ip_address(self, request):
+        ip = super().get_ip_address(request)
         x_forwarded_for = request.headers.get(self.HEADER_NAME, None)
         if x_forwarded_for:
             # HTTP_X_FORWARDED_FOR can be a comma-separated list of IPs.
@@ -130,6 +90,7 @@ class XForwardedForMiddleware:
 
             try:
                 ipaddress.ip_address(client_ip)
+                ip = client_ip
             except ValueError:
                 # This should be OK to log since this is a spoofed address
                 log.warning(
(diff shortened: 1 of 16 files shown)


davidfischer:
Just a note: there's a lot of other [fields](https://developers.cloudflare.com/ruleset-engine/rules-language/fields#dynamic-fields) we can use from Cloudflare.

adserver/middleware.py
```diff
@@ -1,6 +1,93 @@
...
+        if not ip:
+            ip = request.META.get("REMOTE_ADDR", "")
+
+        return ip
+
+    def _get_country(self, request):
+        country_code = request.headers.get(self.COUNTRY_HEADER, None)
+        if country_code == "XX":
+            # CF always adds the header and sets "XX" if it can't geolocate
+            # CF uses "T1" for Tor traffic
```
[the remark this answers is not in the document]
davidfischer: Well, nothing will ever match it for targeting. I suppose somebody could explicitly target Tor but that doesn't seem likely.

config/settings/base.py
```diff
@@ -77,7 +77,8 @@
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
-    "adserver.middleware.XForwardedForMiddleware",
+    "adserver.middleware.ServerInfoMiddleware",
+    "adserver.middleware.CloudflareMiddleware",
```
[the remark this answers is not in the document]
davidfischer: For GeoIP purposes, this is a fine default since it falls back to the GeoIP databases. For getting the true IP header, this isn't safe although the x-forwarded-for header is also unsafe to rely on unless deployed properly. Maybe these should be split into two middlewares. Regardless, we could just have an envvar like `ADSERVER_EXTRA_MIDDLEWARES` and then do something like:

```python
for middleware in ADSERVER_EXTRA_MIDDLEWARES:
    MIDDLEWARE.append(middleware)
```

adserver/views.py
```diff
@@ -599,16 +599,12 @@ def ignore_tracking_reason(self, request, advertisement, offer):
...
-        metro_code = None
-        geo_data = get_geolocation(ip_address)
-        if geo_data:
-            # One or more of these may be None which is OK
-            # Ads targeting countries/regions/metros won't be counted
-            country_code = geo_data["country_code"]
-            region_code = geo_data["region"]
-            metro_code = geo_data["dma_code"]
+        # One or more of these may be None which is OK
+        # Ads targeting countries/regions/metros won't be counted
```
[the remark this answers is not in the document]
davidfischer: If any of these are None, nothing will ever match them. Even the country can be none if somebody spoofs their IP address with something bogus although then they won't get the response.

adserver/utils.py
```diff
@@ -249,22 +252,29 @@ def is_proxy_ip(ip):
...
         try:
-            return geoip.city(ip_address)
+            geolocation = geoip.city(ip_address)
         except AddressNotFoundError:
             log.debug("Could not get geolocation for %s", ip_address)
         except GeoIP2Exception:
             log.warning("Geolocation configuration error")
 
-    return None
+    return geolocation
```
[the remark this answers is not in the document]
davidfischer: Ya, this doesn't quite seem right. Here's how I think it should work:

- If we get a country from Cloudflare, use that.
- Use the GeoIP database just to fill in the region/dma. Maybe remove this since we aren't using it at all?
- We probably don't need to validate the IP address at all here. We should assume that the request's IP is accurate from a middleware.

adserver/tests/test_middleware.py
```diff
@@ -1,34 +1,94 @@
+from django.contrib.auth import get_user_model
 from django.test import TestCase
+from django_dynamic_fixture import get
 
 
 class TestMiddleware(TestCase):
-    def test_xforwarded_for_middleware(self):
+    def test_server_middleware(self):
+        response = self.client.get("/")
+        self.assertFalse("X-Server" in response)
```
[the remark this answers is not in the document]
davidfischer: I changed this to always send this header.

adserver/utils.py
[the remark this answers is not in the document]
davidfischer: I updated this to just always use what the middleware sets.

config/settings/base.py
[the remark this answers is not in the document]
davidfischer: I refactored this a bit. There's now two settings:
- `ADSERVER_GEOIP_MIDDLEWARE`
- `ADSERVER_IPADDRESS_MIDDLEWARE`

These append middleware to the settings. There are provided middlewares for GeoIP databases, Cloudflare GeoIP, Cloudflare Real IP, and X-Forwarded-For.

adserver/middleware.py
```diff
@@ -137,8 +99,110 @@ def _get_ip_address(self, request):
...
+    def get_ip_address(self, request):
+        ip = super().get_ip_address(request)
+        cf_ip = request.headers.get(self.IP_HEADER, None)
+
+        if cf_ip:
+            try:
+                ipaddress.ip_address(cf_ip)
+                ip = cf_ip
+            except ValueError:
+                # This should be OK to log since this is a spoofed address
```
ericholscher, line 129: 👍

adserver/middleware.py
```diff
@@ -137,8 +99,110 @@ def _get_ip_address(self, request):
...
+    SECURITY NOTE: This middleware *SHOULD BE DISABLED* if not running on Cloudflare.
+        Otherwise these headers are spoofable and could result in invalid traffic.
+        Also, Cloudflare IP Geolocation must be enabled.
+    """
+
+    COUNTRY_HEADER = "CF-IPCountry"
+    PROVIDER_NAME = "Cloudflare"
+
+    def get_geoip(self, request):
+        geo = super().get_geoip(request)
```
ericholscher, line 188: This is basically just getting an empty `GeolocationData` instance? Or are there cases where it might already have some data set?

adserver/tests/test_decision_engine.py
```diff
@@ -230,23 +230,23 @@ def test_flight_geo_targeting(self):
...
 
-        self.backend.country_code = "AZ"
+        self.backend.geolocation = GeolocationData("AZ")
         ad, _ = self.backend.get_ad_and_placement()
         self.assertIsNone(ad)
 
-        self.backend.country_code = "RANDOM"
+        self.backend.geolocation = GeolocationData("T1")
         ad, _ = self.backend.get_ad_and_placement()
         self.assertEqual(ad, self.advertisement2)
```
ericholscher, line 251: So we'll serve an ad to Tor users if a flight has only exclude countries, but nothing included?

adserver/utils.py
```diff
@@ -31,13 +31,23 @@
...
-)
-
-
 # Put this here so we don't reload it on each call
 COUNTRY_DICT = dict(countries)
+COUNTRY_DICT["T1"] = "Tor"
+
+
+@dataclass
+class GeolocationData:
```
ericholscher, line 40: I like the dataclass approach 👍

adserver/utils.py
```diff
@@ -31,13 +31,23 @@
...
 
 
-GeolocationTuple = namedtuple(
-    "GeolocationTuple", ["country_code", "region_code", "metro_code"]
-)
-
-
 # Put this here so we don't reload it on each call
 COUNTRY_DICT = dict(countries)
+COUNTRY_DICT["T1"] = "Tor"
```
ericholscher, line 36: I think this probably won't be included in any of our aggregate reports currently, if it isn't in the `regiontopics` listing. Probably fine though..

adserver/utils.py
```diff
@@ -87,14 +97,18 @@ def get_client_ip(request):
...
-    if ip:
-        return ip
+    if hasattr(request, "ip_address"):
+        return request.ip_address
 
-    return request.META.get("REMOTE_ADDR", "")
+    log.warning(
+        "No IP address set by middleware (see CloudflareIpAddressMiddleware). "
+        "Consider enabling an IP address middleware so IPs can be used for ad targeting."
+    )
```
ericholscher, line 109: Nice warning 👍 

Though it will still use the `REMOTE_ADDR` to target, so maybe this message is a bit off?

adserver/views.py
```diff
@@ -650,19 +647,17 @@ def ignore_tracking_reason(self, request, advertisement, offer):
...
             log.log(self.log_level, "Ad impression for unknown publisher")
             reason = "Unknown publisher"
-        elif not advertisement.flight.show_to_geo(
-            country_code, region_code, metro_code
-        ):
+        elif not advertisement.flight.show_to_geo(geo_data):
             # This is very rare but it is visible in ad reports
             # I believe the most common cause for this is somebody uses a VPN and is served an ad
             # Then they turn off their VPN and click on the ad
             log.log(
```
ericholscher, line 654: I've been enjoying structlog on the RTD side. Might be worth looking into, as it lets you `structlog.bind` a set of variables that carry over into all later logs in that request, which leads to some nice outcomes.

config/settings/base.py
```diff
@@ -456,3 +455,16 @@
 
 with open(os.path.join(BASE_DIR, "package.json"), encoding="utf-8") as fd:
     ADSERVER_VERSION = json.load(fd)["version"]
+
+# Additional middleware for setting the user's real IP
+# and translating the IP into a geo for ad targeting.
+ADSERVER_IPADDRESS_MIDDLEWARE = env("ADSERVER_IPADDRESS_MIDDLEWARE", default=None)
+ADSERVER_GEOIP_MIDDLEWARE = env("ADSERVER_GEOIP_MIDDLEWARE", default=None)
```
ericholscher, line 462: Need to remember to set these on deploy.

ericholscher (commented):
This refactor got large! I think it all looks great, but definitely a pretty big change now, so something we need to watch closely.

adserver/middleware.py
davidfischer: It should always be empty.

adserver/tests/test_decision_engine.py
davidfischer: That's correct as written. Maybe add it to the exclusion list?

adserver/utils.py
davidfischer: I don't think we ever want it anywhere. I am just noting it for completeness.

adserver/utils.py
davidfischer: I think it's a pretty important warning to raise. Without some middleware `REMOTE_USER` is almost certain to be wrong. It's probably just a load balancer's IP.

config/settings/base.py
davidfischer: I think the deploy this goes out in will be this and nothing else =)

davidfischer:
You got a review in before I wrote this comment but I did a pretty significant refactoring (as you noticed):

- The middlewares are no longer included by default
- There's new settings (envvars) for including optional middleware
- Separate real IP address middleware from GeoIP middleware. There are middlewares for different approaches (x-forwarded-for vs Cloudflare)
- This approach is easily extended if we ever want to actually target by say 100km of a lat/lng

adserver/views.py
davidfischer: I've never used it but it looks like it has some nice features. I have a branch I've been toying with that does some test only changes and that might be a good place to try out some logging changes.

config/settings/base.py
davidfischer: I wonder if we should set the defaults to the base classes `GeoIpMiddleware` and `IpAddressMiddleware`. That way we could guarantee that `request.ip_address` and `request.geo` will always be present.

davidfischer: I made these the default.

PR #513, opened 2022-02-02
Fix for incorrectly creating new advertisers

davidfischer:
The django-stripe integration is only 1 way. I don't know why I thought it was 2 way.


ericholscher (approved):
Interesting. Would be cool if it was also an API client and 2-way sync.

PR #515, opened 2022-02-04
Connect flights to invoices

davidfischer:
This connects flights to invoices with a M2M. Also, when creating draft invoices through the django admin, the resulting finalized invoice will be connected to the flight. These invoices are shown in the django admin and in the advertiser UI.

### Advertiser UI
Screenshot from 2022-02-04 13-34-39

### Django admin
Screenshot from 2022-02-04 13-34-50


adserver/models.py
```diff
@@ -797,6 +809,11 @@ def ctr(self):
         views = self.total_views
         return calculate_ctr(clicks, views)
 
+    @cached_property
```
ericholscher, line 812: Is there a timeout on this cache? Seems like this might cache for as long as the server is running? Not sure if we need the cache if it's just a DB call?

ericholscher (approved):

adserver/models.py
davidfischer: It doesn't actually use the Django cache framework (that is, memcached/redis). Instead it just caches it for the lifetime of the object. The main benefit is that something like the following only hits the DB once as opposed to twice:

```python
if flight.active_invoices:
    for invoice in flight.active_invoices:
        pass
```

The [docs](https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.functional.cached_property) and [source code](https://docs.djangoproject.com/en/2.2/_modules/django/utils/functional/#cached_property) are pretty short if you've never used it before.

PR #516, opened 2022-02-04
Add a campaign inline to the advertiser admin

davidfischer:


ericholscher (approved):
I've needed this a few times 👍

PR #517, opened 2022-02-08
Disable a publisher completely

davidfischer:
This is currently branched from #515 due to migration numbering. A decision for a disabled publisher returns a 400 bad request.


ericholscher (approved):
Easy enough -- thanks for tackling this 👍

PR #519, opened 2022-02-08
Handle a bug with a forced ad but mismatched ad type

davidfischer:
I found this issue in [Sentry](https://sentry.io/organizations/read-the-docs/issues/2998521932/events/?project=1846644). This can happen when you force a specific ad (rare, usually only in debug) but the ad types don't match the requested placements.


ericholscher (approved):
This seems reasonable, but I liked your previous suggestion of just always forcing the ad regardless of any targeting concerns. I guess this isn't specifically targeting, but a similar outcome.

PR #520, opened 2022-02-08
Add a Python data import script

ericholscher:
This is pretty heavily tied to our integration with the PSF,
but there isn't a better repo to this code in.
We should probably consider having some kind of private "addons" repo,
but for now I at least kept the URL behind an env var just in case.

### DIFF
--- a/adserver/importers/psf.py
+++ b/adserver/importers/psf.py
@@ -127,6 +127,7 @@ def run_import(sync=False, images=False):
 
         valid_ads.add(ad)
 
+    # Disable ads no longer returned by the API
     for iterated_ad in Advertisement.objects.filter(
         flight__campaign__advertiser__slug="psf", live=True
     ):
--- a/adserver/importers/psf.py
+++ b/adserver/importers/psf.py
@@ -25,6 +25,7 @@ def run_import(sync=False, images=False):
     * PyPI Sponsors
     * PSF Sponsors
     * PSF Jobs
+    * PSF Downloads
     * Pycon Sponsors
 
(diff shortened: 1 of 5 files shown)


ericholscher:
@davidfischer this is still a WIP, after talking a bit with the PSF, so no rush here.

ericholscher:
@davidfischer this is ready for review now -- the failing test is coverage, which just means I need to add more tests to this code, which I'll try to do today.

adserver/importers/psf.py
```diff
@@ -0,0 +1,141 @@
...
+                    slug=slugify(name),
+                    flight=flight,
+                )
+            except Advertisement.DoesNotExist:
+                log.warning(f"Failed to get ad for {name}")
+                continue
+
+        valid_ads.add(ad)
+
+    for iterated_ad in Advertisement.objects.filter(
```
davidfischer, line 130:
```suggestion
    # Disable ads no longer returned by the API
    for iterated_ad in Advertisement.objects.filter(
```

davidfischer (approved):
There's a lot of highly custom code specific to the PSF here. That's fine but this kind of code does tend to lose sync with the rest of the codebase over time. I don't know an amazing way to prevent that but maybe at least a testcase that does a dry run or something?

ericholscher:
I'm going to merge this, since I think we'll be actively running this so it shouldn't lose sync. If the API changes on their side or our side, it will start showing up in Sentry as WARNING, and we should be able to handle it then. I don't know if a dry run test will really do much unfortunately :/

PR #521, opened 2022-02-09
Fix typo with keyword aggregation

davidfischer:
Fixes a bug with keyword aggregation introduced in version v0.44.0 (January 26, 2022). All aggregations since then are wrong and will need to be re-run after this is merged.

I had verified that the query generated was correct but didn't verify the actual writing to the database. That has been corrected with a unit test.

### DIFF
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -319,7 +319,7 @@ def daily_update_keywords(day=None):
                 decisions=F("decisions") + values["total_decisions"],
                 offers=F("offers") + values["total_offers"],
                 views=F("views") + values["total_views"],
-                clicks=F("clicks") + values["total_views"],
+                clicks=F("clicks") + values["total_clicks"],
             )
 
 
--- a/adserver/tests/test_tasks.py
+++ b/adserver/tests/test_tasks.py
@@ -5,9 +5,11 @@ from django_dynamic_fixture import get
 from django_slack.utils import get_backend
 
 from ..models import AdImpression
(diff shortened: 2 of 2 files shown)


adserver/tests/test_tasks.py
```diff
@@ -70,6 +72,79 @@ def test_calculate_publisher_ctrs(self):
         self.publisher.refresh_from_db()
         self.assertEqual(self.publisher.sampled_ctr, 20)
 
+    def test_daily_update_keywords(self):
+        # Keyword Aggregation requires some targeting
+        self.flight.targeting_parameters = {
+            "include_keywords": ["django"],
```
MxRaCohen, line 78: Given the way we sum, would the result be changed if the targeting was `"python"`?

adserver/tasks.py
```diff
@@ -319,7 +319,7 @@ def daily_update_keywords(day=None):
                 decisions=F("decisions") + values["total_decisions"],
                 offers=F("offers") + values["total_offers"],
                 views=F("views") + values["total_views"],
-                clicks=F("clicks") + values["total_views"],
+                clicks=F("clicks") + values["total_clicks"],
```
ericholscher, line 322: Whoops 📈

ericholscher (approved):

adserver/tests/test_tasks.py
davidfischer: We generate a different record for each one. Since all the ads were for both, the data will be duplicated.

PR #525, opened 2022-02-10
Add an existing user to an advertiser

davidfischer:
This is just a QoL improvement that simplifies the workflow of adding an existing user.


ericholscher (approved):
Seems simple enough. Probably a useful improvement for publisehrs as well, but if that's the case I'll often just add them in the admin.

PR #526, opened 2022-02-10
Put the priority multiplier in the flight form

davidfischer:


ericholscher (approved):
I know this was often something that needed back & forth, so easy enough to expose 👍

PR #530, opened 2022-02-15
Record data for forced ads if they are unpaid.

ericholscher:
This is required for the PSF sponsor use case,
and I think makes sense with never billing for them..


davidfischer (approved):
A test for this seems warranted but otherwise looks good.

ericholscher:
@davidfischer Added. This would be good to deploy if we can while I'm gone, since currently these aren't being counted.

PR #532, opened 2022-02-17
Aggregation task performance improvements

davidfischer:
This adds performance improvements to our daily aggregations. This builds on the improvements in https://github.com/readthedocs/ethical-ad-server/pull/508 and uses similar ones for the other aggregations. This should cut down both our read and write queries by a factor of ~4. I expect this to cut the time taken to do a daily aggregation by 50-66% or so.

Notably, the tests added in this PR work with or without the changes to `tasks.py`.

### DIFF
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -12,11 +12,7 @@ from django.utils.timezone import is_naive
 from django.utils.timezone import utc
 from django_slack import slack_message
 
-from .constants import CLICKS
-from .constants import IMPRESSION_TYPES
-from .constants import OFFERS
 from .constants import PAID_CAMPAIGN
-from .constants import VIEWS
 from .models import AdImpression
 from .models import Flight
 from .models import GeoImpression
@@ -67,26 +63,6 @@ def _get_day(day):
     return (start_date, end_date)
 
 
-def _default_filters(impression_type, start_date, end_date):
-    """Filter the queryset by date and impression type."""
-    # Use the replica for this query, since it's how we do all our reporting queries,
-    # and it currently hammers the prod DB.
-    queryset = Offer.objects.using(settings.REPLICA_SLUG).filter(
-        date__gte=start_date,
-        date__lt=end_date,  # Things at UTC midnight should count towards tomorrow
-        # is_refunded=False,  # This causes the query to be a filtered index and is much slower
-    )
-
-    if impression_type == CLICKS:
-        queryset = queryset.filter(clicked=True)
-    elif impression_type == VIEWS:
-        queryset = queryset.filter(viewed=True)
-    elif impression_type == OFFERS:
-        queryset = queryset.filter(advertisement__isnull=False)
-
-    return queryset
-
-
 @app.task()
 def daily_update_geos(
     day=None, geo=True, region=True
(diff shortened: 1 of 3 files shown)


adserver/tasks.py
```diff
@@ -114,71 +90,96 @@ def daily_update_geos(
...
-                ]
+    topic_mapping = defaultdict(
+        lambda: {
+            "decisions": 0,
+            "offers": 0,
+            "views": 0,
+            "clicks": 0,
+        }
+    )
+    queryset = Offer.objects.using(settings.REPLICA_SLUG).filter(
```
ericholscher, line 101: Might be good to keep all these queries the same, where we're just doing a single filter on `Offer` as done below: https://github.com/readthedocs/ethical-ad-server/pull/532/files#diff-bdef58a59b4162498b09457b29f37148cb2494ada036cc9ba8293f625b14b1bdR196-R214

adserver/tasks.py
```diff
@@ -227,26 +239,36 @@ def daily_update_impressions(day=None):
...
+            date__gte=start_date,
+            date__lt=end_date,  # Things at UTC midnight should count towards tomorrow
+        )
+        .values("publisher", "advertisement")
+        # This needs to be publisher and not advertisement to gets decisions properly
+        .annotate(
+            total_decisions=Count("publisher"),
+            total_offers=Count("publisher", filter=Q(advertisement__isnull=False)),
+            total_views=Count("publisher", filter=Q(viewed=True)),
+            total_clicks=Count("publisher", filter=Q(clicked=True)),
```
ericholscher, line 254: I wonder about some way to pull this query into a helper function? We'd probably need to pass the `values` and `primary_value` (which would be `publisher` in this case). That way it reduces errors around filtering differently. 

All of the queries tend to be slightly different though, so maybe it doesn't make sense. I think there's a reason I ended up with `_default_filters` instead of a larger helper function. 🤷

adserver/tasks.py
```diff
@@ -341,90 +363,115 @@ def daily_update_regiontopic(day=None):  # pylint: disable=too-many-branches
...
+        elif country in eu_aus_nz:
+            region = "eu-aus-nz"
+        elif country in wider_apac:
+            region = "wider-apac"
+        elif country in latin_america:
+            region = "latin-america"
+        elif country in africa:
+            region = "africa"
+        else:
+            region = "global"
```
ericholscher, line 445: I think these regions aren't in sync with the other logic around regions, which includes `south-asia`. Another problem with this file not being very DRY (that I introduced).

ericholscher (approved):
Lots of good updates here. I'd love to simplify this code if possible to reduce bugs/oversights, but not sure I have a great solution. I think the real answer is having fewer indexes, but we seem to be using all of these in various places.

davidfischer:
> Lots of good updates here. I'd love to simplify this code if possible to reduce bugs/oversights, but not sure I have a great solution. I think the real answer is having fewer indexes, but we seem to be using all of these in various places.

I tend to agree. I'm not sure what the path forward is here though. Each of these indexes and queries is slightly different.

ericholscher:
💯

PR #533, opened 2022-02-17
Performance improvement to offer recording

davidfischer:
This is a small performance optimization to recording offers. It records the offer and decision at the same time. This should cut down by a read query and a write query.


ericholscher (approved):
Solid speedup. Definitely something we're doing quite often, so nice to be able to write them both at once.

PR #534, opened 2022-02-22
Force IP Geolocation if there's a passed userip

davidfischer:


adserver/tests/test_api.py
```diff
@@ -1165,6 +1165,27 @@ def test_ad_click_and_tracking(self):
...
+            get_geo.return_value = GeolocationData("CA")
+
+            resp = self.client.post(
+                self.url, json.dumps(data), content_type="application/json"
+            )
+        self.assertEqual(resp.status_code, 200, resp.content)
+
+        # Check that the new IP was used for ad targeting
+        self.assertEqual(resp["X-Adserver-RealIP"], new_ip)
+        self.assertEqual(resp["X-Adserver-Country"], "CA")
```
agjohnson, line 1187: Without this PR, this test would give a `X-Adserver-RealIP` header of `8.8.8.8`, correct?

agjohnson (approved):
Looks good!

adserver/tests/test_api.py
davidfischer: Actually, the `RealIP` would be the same but the country wouldn't be correct.

PR #536, opened 2022-02-23
Flight renewal form

davidfischer:
- Copy existing ads (you can choose them)
- Update start and end date
- Uses same targeting
- Currently the view is staff only

## Screenshot
Screenshot from 2022-02-23 11-52-29

### DIFF
--- a/adserver/forms.py
+++ b/adserver/forms.py
@@ -1,5 +1,6 @@
 """Forms for the ad server."""
 import logging
+from datetime import timedelta
 
 import bleach
 import stripe
@@ -20,6 +21,7 @@ from django.core.mail import EmailMessage
 from django.db.models import Q
 from django.template.loader import render_to_string
 from django.urls import reverse
+from django.utils import timezone
 from django.utils.crypto import get_random_string
 from django.utils.html import format_html
 from django.utils.text import slugify
@@ -48,6 +50,11 @@ from .validators import TargetingParametersValidator
 
 log = logging.getLogger(__name__)  # noqa
 User = get_user_model()
+
+
+class AdvertisementMultipleChoiceField(forms.ModelMultipleChoiceField):
+    def label_from_instance(self, obj):
+        return obj
 
 
 class FlightMixin:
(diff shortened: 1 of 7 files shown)


adserver/forms.py
```diff
@@ -50,6 +52,11 @@
 User = get_user_model()
 
 
+class AdvertisementMultipleChoiceField(forms.ModelMultipleChoiceField):
+    def label_from_instance(self, obj):
+        return obj
```
ericholscher, line 57: This seems like an odd override class, probably worth a comment or docstring.

adserver/forms.py
```diff
@@ -398,6 +409,129 @@ class Meta:
...
+                css_class="my-3",
+            ),
+            Submit("submit", _("Create flight via renewal")),
+        )
+        return helper
+
+    def save(self, commit=True):
+        assert commit, "Delayed saving is not supported on this form"
+
+        instance = super().save(commit=True)
```
ericholscher, line 496: Should be using the `commit` variable here, even though we know it's truthy.

adserver/templates/adserver/includes/widgets/advertisement-form-option.html
```diff
@@ -0,0 +1,50 @@
+{# Adapted from bootstrap4/layout/checkboxselectmultiple.html in crispy forms #}
```
ericholscher, line 1: Not 100% sure how to review this. Might be useful to note what was changed here (eg. `Changes on lines 6-9`)

adserver/views.py
```diff
@@ -358,6 +359,53 @@ def get_success_url(self):
...
+        return super().dispatch(request, *args, **kwargs)
+
+    def form_valid(self, form):
+        result = super().form_valid(form)
+        flight = self.object
+        messages.success(
+            self.request,
+            _("Successfully created new flight '%(flight)s' via renewal")
+            % {"flight": flight},
+        )
```
ericholscher, line 385: Do we want to use f-strings here?

ericholscher (approved):
This looks like a great enhancement! A few small nits, but excited to see this one go out.

adserver/templates/adserver/includes/widgets/advertisement-form-option.html
davidfischer: There are A LOT of changes. This file is now specific to advertisements.

PR #537, opened 2022-02-25
Post to Slack when an invoice is paid

davidfischer:
This will post a Slack message when an invoice is paid.

## Testing

* Set `STRIPE_SECRET_KEY` and `STRIPE_TEST_SECRET_KEY` in `.envs/local/django`.
* Follow instructions here: https://dashboard.stripe.com/test/webhooks/create?endpoint_location=local
* Test the hook with `stripe trigger invoice.payment_succeeded`

### DIFF
--- a/adserver/apps.py
+++ b/adserver/apps.py
@@ -8,3 +8,4 @@ class AdserverConfig(AppConfig):
 
     def ready(self):
         import adserver.tasks  # noqa
+        import adserver.hooks  # noqa
--- a/adserver/hooks.py
+++ b/adserver/hooks.py
@@ -1,0 +1,32 @@
+"""Handle specific Stripe webhooks with special logic."""
+import logging
+
+import stripe
+from django_slack import slack_message
+from djstripe import webhooks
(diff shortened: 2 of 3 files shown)


adserver/apps.py
```diff
@@ -8,3 +8,4 @@ class AdserverConfig(AppConfig):
 
     def ready(self):
         import adserver.tasks  # noqa
+        import adserver.hooks  # noqa
```
ericholscher, line 11: I wonder about calling this `signals` instead? It seems that it's using the same pattern, and would be where folks might look for similar logic in a Django codebase. I don't feel strongly, but just a thought.

ericholscher (approved):

adserver/apps.py
davidfischer: This isn't a signal. This is a django-stripe specific hook. I realize this isn't the standard place but there isn't exactly a standard place to put it. I think `webhooks.py` is better than `signals.py` overall. Since it isn't precisely a true webhook either, I chose `hooks.py`.

PR #539, opened 2022-03-01
Django 3.2 upgrade

davidfischer:
This ended up not being so bad. The tests are passing, the migrations seem minimal, and everything looks like it's working.


davidfischer:
Spoke too soon due to some issues with cached dependencies locally. I'm glad CI caught this.

stsewd (approved):
doesn't look like this required any mayor changes

PR #540, opened 2022-03-02
Close flights when complete

davidfischer:


ericholscher (approved):
This looks good. I do wonder about adding the ability to email the users in this case, while we're changing up how things work?

davidfischer:
> This looks good. I do wonder about adding the ability to email the users in this case, while we're changing up how things work?

I think we should start doing that as more of this process gets automated.

PR #546, opened 2022-03-17
Add view_time to AdImpression model

ericholscher:
This will allow us to start showing it in reports and in various other places.

This just adds indexing, and doesn't display the data anywhere yet. But it will allow us to start looking at it in Metabase at least, and then expand it into user reports.


adserver/admin.py
```diff
@@ -969,7 +973,7 @@ class RegionTopicAdmin(admin.ModelAdmin):
 
 # Don't register Impression Admin's outside dev, since they will just 502 from too much data.
 if settings.DEBUG:
-    admin.site.register(AdImpression, AdImpressionsAdmin)
+    admin.site.register(AdImpression, ImpressionAdmin)
     admin.site.register(UpliftImpression, AdImpressionsAdmin)
```
davidfischer, line 977: Maybe if we're going to use a different admin from `UpliftImpression`, we should just change the name of that one. eg. AdImpressionsAdmin -> UpliftImpressionAdmin, AdImpression -> AdImpressionAdmin.

davidfischer (approved):
Looks good. Hopefully the migration isn't a problem (it shouldn't be).

PR #552, opened 2022-03-24
Adds a Front (front.com) email backend

davidfischer:
Adds a Front email backend to the project. This allows sending messages through Front or creating drafts as easily as sending a regular email through Django. This could be set as the default [EMAIL_BACKEND](https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-EMAIL_BACKEND) but it doesn't have to be and could be used as a secondary backend almost as easily.

In the future, we could consider breaking this out into a separate open source project and using it on RTD.

## Testing

You'll need to set `FRONT_CHANNEL`, `FRONT_AUTHOR` (for drafts), and `FRONT_TOKEN` in `.envs/local/django`.

Then, can be tested with the following:

```python
from django.core import mail

with mail.get_connection("frontbackend.backend.EmailBackend") as connection:
    message = mail.EmailMessage(
        "Test subject",
        "This is a test body.", 
        "from@example.com", 
        ["to@example.com"],
        connection=connection,
    )

    # Optionally make it a draft
    message.draft = True

    # Override archiving a message
    message.archive = False

    message.send()
```

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -188,6 +188,7 @@ DEFAULT_FROM_EMAIL = SERVER_EMAIL
 EMAIL_TIMEOUT = 5
 
 # For sending email through Front.
+FRONT_BACKEND = "frontbackend.backend.EmailBackend"
 FRONT_TOKEN = env("FRONT_TOKEN", default=None)
 FRONT_CHANNEL = env("FRONT_CHANNEL", default=None)
 FRONT_AUTHOR = env("FRONT_AUTHOR", default=None)
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -187,10 +187,13 @@ SERVER_EMAIL = "noreply@ethicalads.io"
 DEFAULT_FROM_EMAIL = SERVER_EMAIL
 EMAIL_TIMEOUT = 5
 
-# For sending Front email. Only used for Payouts.
+# For sending email through Front.
 FRONT_TOKEN = env("FRONT_TOKEN", default=None)
 FRONT_CHANNEL = env("FRONT_CHANNEL", default=None)
 FRONT_AUTHOR = env("FRONT_AUTHOR", default=None)
+FRONT_SENDER_NAME = env("FRONT_SENDER_NAME", default=None)
+FRONT_ARCHIVE = env.bool("FRONT_ARCHIVE", default=False)
+
 
 # Internationalization
 # https://docs.djangoproject.com/en/3.2/topics/i18n/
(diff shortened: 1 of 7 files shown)


config/settings/base.py
```diff
@@ -187,10 +187,13 @@
 DEFAULT_FROM_EMAIL = SERVER_EMAIL
 EMAIL_TIMEOUT = 5
 
-# For sending Front email. Only used for Payouts.
+# For sending email through Front.
 FRONT_TOKEN = env("FRONT_TOKEN", default=None)
 FRONT_CHANNEL = env("FRONT_CHANNEL", default=None)
 FRONT_AUTHOR = env("FRONT_AUTHOR", default=None)
+FRONT_SENDER_NAME = env("FRONT_SENDER_NAME", default=None)
```
ericholscher, line 194: Should this default to `DEFAULT_FROM_EMAIL`?

frontbackend/backend.py
```diff
@@ -0,0 +1,127 @@
...
+
+        if not token or not channel:
+            raise NotImplementedError(
+                "For the Front email backend, "
+                "settings.FRONT_TOKEN and settings.FRONT_CHANNEL "
+                "must be set."
+            )
+
+        # Whether to archive messages on send
+        self.archive = settings.FRONT_ARCHIVE
```
ericholscher, line 51: Seems like we should be able to set this per message, right? The current implementation allows the user to choose whether to archive it which is useful if you want to follow up on it. Probably not super important though.

 https://github.com/readthedocs/ethical-ad-server/blob/192f6c6af45a740f140f65e9d308c88b04897627/adserver/staff/forms.py#L352

frontbackend/settings.py
```diff
@@ -0,0 +1,29 @@
+"""Default settings for the Front email backend."""
+from django.conf import settings
```
ericholscher, line 2: Not quite sure why we need to duplicate the settings code? Could be useful to add in a comment if there's a design reason for it.

ericholscher (commented):
This looks good. It doesn't currently integrate it into the codebase, is that the next step?

config/settings/base.py
davidfischer: This actually isn't an email address. It is just a display name. The "from" address will always be configured by what the address is from `FRONT_CHANNEL`. For payouts, this is set to "EthicalAds by Read the Docs". I think that's a reasonable value.

frontbackend/backend.py
davidfischer: I guess it's easy enough to make this configure the default and allow some way to override it.

frontbackend/settings.py
ericholscher: Ah, gotcha. So just a default value if it's used in another project 👍

ericholscher (approved):
Looks good as an initial backend

PR #556, opened 2022-03-28
Send a flight wrapup email

davidfischer:
~~**NOTE:** this is not quite ready for merging. I would like to write some tests, flesh out the actual email a bit more, and generally clean things up before merging.~~

This change will send one email to every authorized user of an advertiser (unless they opt out of notification settings) each day that a flight ends. The email will have information on all flights that end that day. It's somewhat common for an advertiser to have different flights end the same day -- especially if they do a content or geographic split.

Currently, the code will just create a draft email that an individual sales person can send.

The goals of this email are:

- Show details about the flights and how they performed
- Get in touch with sales

## Future possibilities
These are things I don't want to do in this PR but could be good improvements:

- Use a styled HTML email template. While I generally dislike styled HTML email for cold/warm emails, for a transactional email about a user's account, sending better looking emails shows polish. I think it's a good idea.
- Include graphs or other details from metabase. I'm not sure this is a possibility but it would be nice.
- Warn on overdue invoices


adserver/tasks.py
```diff
@@ -581,6 +590,57 @@ def notify_of_completed_flights():
             flight.live = False
             flight.save()
 
+            completed_flights_by_advertiser[flight.campaign.advertiser.slug].append(
+                flight
+            )
+
+    # Send notification to advertiser - one email even if multiple flights finished
+    if settings.FRONT_TOKEN and settings.FRONT_AUTHOR and settings.FRONT_CHANNEL:
```
ericholscher, line 598: Seems like we should probably have this logic in one place, and then `settings.FRONT_ENABLED` as an alias?

adserver/tasks.py
```diff
@@ -581,6 +590,57 @@ def notify_of_completed_flights():
...
+                    if f.state == FLIGHT_STATE_UPCOMING
+                ],
+            }
+
+            with mail.get_connection(settings.FRONT_BACKEND) as connection:
+                message = mail.EmailMessage(
+                    _("Advertising flight wrapup - %(name)s") % {"name": site.name},
+                    render_to_string("adserver/email/flight_wrapup.html", context),
+                    settings.SERVER_EMAIL,  # Front doesn't use this
+                    to_addresses,
```
ericholscher, line 635: I always prefer kwargs if possible on these, so we know what these arguments are.

adserver/templates/adserver/email/base.html
```diff
@@ -0,0 +1,18 @@
+{# Email intentionally uses an old doctype for maximum compatibility #}
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "https://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<html xmlns="https://www.w3.org/1999/xhtml">
+<head>
+  {# This title is never displayed but there's a block in case you want to customize it #}
```
ericholscher, line 5: Don't quite follow this. Will it be displayed if it's customized?

adserver/templates/adserver/email/flight_wrapup.html
```diff
@@ -0,0 +1,24 @@
+{% extends "adserver/email/base.html %}
+{% load i18n %}
+
+
+<p>{% blocktrans %}Hi,{% endblocktrans %}</p>
+
+<p>{% blocktrans with site_name=site.name %}Thank you for advertising with {{ site_name }}!{% endblocktrans %}</p>
+
+<p>{% blocktrans %}Your campaign haa wrapped up.{% endblocktrans %}</p>
```
ericholscher, line 9:
```suggestion
<p>{% blocktrans %}Your campaign has wrapped up:{% endblocktrans %}</p>
```

ericholscher (approved):
This looks like a good start. Excited to get this more automated, and really like the draft workflow as a way to roll it out slowly ❤️

adserver/templates/adserver/email/base.html
davidfischer: No, it will not. The `<title>` tag in an HTML is never seen unless somebody looks at the original source of the email.

adserver/templates/adserver/email/flight_wrapup.html
davidfischer: I have a followup commit that fixes this.

davidfischer:
The checks failure is due to some pip versioning issues. I'll prep a fix for it.

PR #560, opened 2022-03-31
Log mismatched clients between offer and impression

davidfischer:


adserver/views.py
```diff
@@ -726,6 +728,16 @@ def ignore_tracking_reason(self, request, advertisement, offer):
             )
             reason = "Ratelimited view impression"
 
+        if offer and offer.client_id != client_id:
+            # Because this block doesn't set a reason, it will only log mismatched clients
+            # Not stop them.
+            log.log(
+                self.log_security_level,
```
ericholscher, line 735: This seems less explicit to me than just `log.warning`. I'm guessing there's a reason it's this way, but it feels a bit less explicit to me.

ericholscher (approved):

adserver/views.py
davidfischer: This makes it easy to change a bunch of security sensitive logs in this section.

PR #564, opened 2022-04-13
Offline keyword and topic analysis

davidfischer:
~~**Note:** this is not quite ready for merging. See the TODO section.~~

Adds offline keyword and topic analysis to improve content targeting of advertising.

## Considerations

- Could there perhaps be an off-the-shelf analyzer we could use? Perhaps something along the lines of Solr/Lucene? It might be overkill to start and we could probably swap over to it over time.
- I expect the database of analyzed URLs to get pretty large although probably much smaller than overall offers.
- I'm not sure how well this will work for single-page apps.

## TODO

* [x] Modeling
* [x] Celery-based HTML parser/analyzer
* [ ] Make the ad decision engine use the keywords (possibly in a separate PR)


adserver/analyzer/models.py
```diff
@@ -0,0 +1,62 @@
...
+        blank=True,
+        help_text=_("Last time the ad server analyzed this URL"),
+    )
+    last_ad_served_date = models.DateTimeField(
+        default=None,
+        null=True,
+        blank=True,
+        help_text=_("Last time an ad was served for this URL"),
+    )
+    visits_since_last_analyzed = models.PositiveIntegerField(
```
ericholscher, line 51: I wonder if we want to actually increment this, or just have a boolean value? I guess the count isn't super hard to get if we want to generate it nightly similar to our indexing.

adserver/analyzer/models.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+
+    publisher = models.ForeignKey(
+        Publisher,
+        help_text=_("Publisher where this URL appears"),
+        on_delete=models.CASCADE,
+    )
+
+    # Fields below are updated by the analyzer
+    keywords = JSONField(_("Keywords for this URL"), blank=True, null=True)
```
ericholscher, line 37: Is there a good way to document the presumed schema for this JSON? Is the model the right place for it?

ericholscher (approved):
Modeling looks good 👍

adserver/analyzer/models.py
davidfischer: I was planning to add a validator.

adserver/analyzer/models.py
davidfischer: This is definitely a part I'm not so sure of. Ideally we don't have a celery task per ad offer. That'll be a lot of tasks. Generating nightly and updating both `last_ad_served_date` and this field seems like a good approach.

davidfischer:
I added a naive analyzer that just does basically what the ad client does -- count keywords. Longer term, I think I'll look at:

* [PKE](https://github.com/boudinfl/pke)
* [Multi-Rake](https://github.com/vgrabovets/multi_rake)
* [Yake](https://github.com/LIAAD/yake)
* Or something similar

Right now, this doesn't actually use this data in the ad decision but I want to run it and sort of see what happens with our data and how fast it runs. Then we can assess the next steps.

adserver/analyzer/backends/base.py
```diff
@@ -0,0 +1,64 @@
...
+
+    """Base class that all analyzers should extend."""
+
+    def __init__(self, url, **kwargs):
+        """Base constructor."""
+        self.url = url
+
+        site = get_current_site(request=None)
+        self.user_agent = (
+            f"EthicalAds Analyzer/{settings.ADSERVER_VERSION} <{site.domain}>"
```
ericholscher, line 23: I do wonder if we should use the standard precursor that all bots seem to, eg:

```
Mozilla/5.0 (compatible; bingbot/2.0; +http://www.bing.com/bingbot.htm)
```

adserver/analyzer/backends/naive.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # CSS selectors to select the "main" content of the page
+    # The first of these to match anything is used
+    MAIN_CONTENT_SELECTORS = (
+        "[role='main']",
+        "main",
+        "body",
+    )
+
+    MAX_WORDS_ANALYZED = 1000
```
ericholscher, line 26: I assume this isn't a performance limit, but that we're just wanting to see what is at the top of the page? We can probably bump this up over time if needed, but probably fine to start with.

adserver/analyzer/tasks.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    normalized_url = normalize_url(url)
+
+    publisher = Publisher.objects.filter(slug=publisher_slug).first()
+    if not publisher:
+        log.warning("Analyzing URL for nonexistent publisher: %s", publisher_slug)
+        return
+
+    backend = get_url_analyzer_backend()(url)
+    keywords = backend.analyze()  # Can be None
```
ericholscher, line 34: I wonder if it should be `[]` instead when no keywords are found? Do `None` and `[]` mean different things in this case?

adserver/analyzer/tasks.py
```diff
@@ -0,0 +1,82 @@
...
+
+@app.task
+def daily_visited_urls(days=2):
+    """Analyze URLs that were visited recently but have never been analyzed."""
+    dt_cutoff = timezone.now() - datetime.timedelta(days=days)
+
+    # Loop over each publisher and query the offers table for distinct URLs
+    for publisher in Publisher.objects.all():
+        for obj in (
+            Offer.objects.filter(publisher=publisher, date__gt=dt_cutoff)
```
ericholscher, line 57: Would it be faster do just query Offers once, and have it return the URL & the publisher?

adserver/analyzer/tasks.py
```diff
@@ -0,0 +1,82 @@
...
+            url = obj["url"]
+            if not url:
+                # Blank URLs can't be analyzed
+                continue
+
+            normalized_url = normalize_url(url)
+            if not AnalyzedUrl.objects.filter(
+                url=normalized_url, publisher_id=publisher.id
+            ).exists():
+                analyze_url.apply_async(args=[normalized_url, publisher.slug])
```
ericholscher, line 70: This will probably blow up our celery queue for a long period of time. Not a huge deal, but it would likely block other celery tasks until it was finished. 

It might be nicer in the future to have a task run per publisher, or a single task that does all this logic, similar to our nightly indexers.

requirements/base.txt
```diff
@@ -58,3 +58,6 @@ dj-stripe==2.5.1
 
 # JWT used by Metabase embedding
 PyJWT==2.1.0
+
+# Used by the keyword/topic analyzer
+beautifulsoup4==4.11.1
```
ericholscher, line 63: I think we're using https://github.com/rushter/selectolax in RTD, mostly for performance reasons. This is fine for now, but worth keeping an eye on if we're spending a lot of time parsing.

adserver/analyzer/tests.py
```diff
@@ -0,0 +1,228 @@
...
+            normalize_url("https://example.com/path"), "https://example.com/path"
+        )
+        self.assertEqual(
+            normalize_url("https://example.com/path?myparam=myval"),
+            "https://example.com/path?myparam=myval",
+        )
+        self.assertEqual(
+            normalize_url("https://example.com/#fragment"),
+            "https://example.com/#fragment",
+        )
```
ericholscher, line 51: We presumably want to strip fragments, since we aren't going to take them into account, and they will add multiple URL entries?

adserver/analyzer/tests.py
```diff
@@ -0,0 +1,228 @@
...
+        self.publisher = get(Publisher)
+        self.analyzed_url = AnalyzedUrl.objects.create(
+            url="https://example.com",
+            publisher=self.publisher,
+            keywords=["python", "django"],
+        )
+
+    def test_str(self):
+        self.assertEqual(
+            str(self.analyzed_url), "['python', 'django'] on https://example.com"
```
ericholscher, line 81: This doesn't seem super useful. I wouldn't expect a test to break if I change the `str` method, I don't think.

adserver/analyzer/tasks.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+@app.task
+def weekly_analyze_urls(days=7):
+    """Analyze URLs that have been visited but not analyzed since DAYS number of days."""
+    dt_cutoff = timezone.now() - datetime.timedelta(days=days)
+
+    analyzed_urls = AnalyzedUrl.objects.filter(
+        last_analyzed_date__lt=dt_cutoff, visits_since_last_analyzed__gt=0
```
ericholscher, line 79: Will `visits_since_last_analyzed` ever be >0? I don't see where it's being incremented in the code.

adserver/analyzer/tasks.py
```diff
@@ -0,0 +1,82 @@
...
+
+    if not created:
+        url_obj.keywords = keywords
+        url_obj.last_analyzed_date = timezone.now()
+        url_obj.visits_since_last_analyzed = 0
+        url_obj.save()
+
+
+@app.task
+def daily_visited_urls(days=2):
```
ericholscher, line 50: I believe we have the celery tasks defined in the `production` settings. Are we not hooking these up to run nightly yet, but manually testing them?

ericholscher (approved):
This is a great first implementation. Excited to get this out and see how it performs in reality.

adserver/analyzer/backends/naive.py
davidfischer: If anything, I'd reduce it. 1000 words is multiple paragraphs. If it doesn't appear in the first 5-10 paragraphs, it's probably not relevant.

adserver/analyzer/tasks.py
davidfischer: Yes, they mean different things. `[]` means no keywords. `None` is a failure of some sort.

adserver/analyzer/tasks.py
davidfischer: I'm not sure. It's worth a test.

adserver/analyzer/tests.py
davidfischer: Without fragments, I think there's no hope for SPAs. I think this may require some experimentation.

adserver/analyzer/tests.py
davidfischer: It isn't super useful except it makes sure the string method works correctly. It should be easy to change if the `str` method changes.

adserver/analyzer/tasks.py
davidfischer: It isn't currently being incremented. We need a task which goes over our `Offers` and increments it.

adserver/analyzer/tasks.py
davidfischer: Correct.

adserver/analyzer/backends/naive.py
ericholscher: Eh, especially with a naive implementation, I feel like all the text is probably useful to analyze, since it will show what is included on the page. If we get a lot smarter about classification we can be pickier, but given the current approach I feel like the more data the better.

adserver/analyzer/tests.py
ericholscher: Ack true... I don't think there's much hope for SPA's anyway, since we aren't using something like headlesschrome to get the URL. That could be a v2, something with JavaScript.

PR #573, opened 2022-05-02
Rework the analysis tasks

davidfischer:
- Count the number of times URLs display paid ads daily.
- (Re-)scan URLs that have 50+ visits and haven't been scanned in more than a week.

This should space out our scanning a bit more and focus it on URLs that get lots of visits.


adserver/analyzer/models.py
```diff
@@ -41,6 +41,7 @@ class AnalyzedUrl(TimeStampedModel):
         blank=True,
         help_text=_("Last time the ad server analyzed this URL"),
     )
+    # This is only accurate to the day
```
ericholscher, line 44: Should it be a `Date` field then?

adserver/analyzer/tasks.py
```diff
@@ -47,36 +53,64 @@ def analyze_url(url, publisher_slug):
...
+            publisher_id=publisher_id,
+            defaults={
+                "last_ad_served_date": timezone.now(),
+                "visits_since_last_analyzed": visits,
+            },
+        )
+
+        if not created:
+            # If the object wasn't created, add the visits and set the ad last served date
+            url_obj.update(
```
ericholscher, line 98: Huh, I thought you could only call `update` on a queryset, not an object.

I wonder if we can this to make it a bit cleaner? https://docs.djangoproject.com/en/3.2/ref/models/querysets/#update-or-create

adserver/analyzer/tasks.py
```diff
@@ -47,36 +53,64 @@ def analyze_url(url, publisher_slug):
...
+                visits_since_last_analyzed=models.F("visits_since_last_analyzed")
+                + visits,
+            )
 
 
 @app.task
-def weekly_analyze_urls(days=7):
-    """Analyze URLs that have been visited but not analyzed since DAYS number of days."""
+def daily_analyze_urls(days=7, min_visits=50):
+    """Analyze URLs that have been visited VISITS times but not analyzed since DAYS number of days."""
```
ericholscher, line 107: Should the kwarg be `visits` to match this docstring? The caps are a bit odd, I'd probably do:

```suggestion
    """Analyze URLs that have been visited `visits` times but not analyzed since `days` number of days."""
```

adserver/utils.py
```diff
@@ -55,6 +59,27 @@ def get_ad_day():
     return timezone.now().replace(hour=0, minute=0, second=0, microsecond=0)
 
 
+def get_day(day=None):
+    """
+    Get the start and end time for a given day.
+
+    Always return two datetimes with timezone.
+    If day is `None`, use today.
```
ericholscher, line 67:
```suggestion
    If `day` is `None`, use today.
```

adserver/analyzer/models.py
davidfischer: It probably should. Yes.

adserver/analyzer/tasks.py
davidfischer: I don't feel that strongly. I do think `min_visits` is a more clear variable name. Maybe just use `min_visits` everywhere?

adserver/analyzer/tasks.py
davidfischer: Oooo, never used that before.

davidfischer: Actually, since I'm using an `F()` query, I don't think I can use this method. I was looking forward to it.

adserver/analyzer/tasks.py
ericholscher: `min_visits` is fine, mostly just commenting on the docstring syntax.

adserver/analyzer/tasks.py
ericholscher: Gotcha. Still confused how we're calling `update` on a object instead of a queryset though, unless I'm reading the code wrong?

adserver/analyzer/tasks.py
davidfischer: Fixed up.

adserver/analyzer/tasks.py
davidfischer: Good catch. This is a problem. Adding a test for this and fixing it.

ericholscher (approved):

PR #579, opened 2022-05-13
Add a daily cap for publishers

davidfischer:
- This is especially useful when evaluating new publishers
- Useful for publishers that get large traffic spikes
- This should only add "time to add" for publishers with the cap


adserver/models.py
```diff
@@ -288,6 +305,20 @@ def payout_url(self):
...
+    def get_daily_earn(self):
+        """Get how much this publisher has earned today."""
+        cache_key = self.daily_earn_cache_key
+        return cache.get(cache_key, default=0.0)
+
+    def increment_daily_earn(self, delta):
+        """Increment how much this publisher has earned today."""
+        cache_key = self.daily_earn_cache_key
+        try:
+            cache.incr(cache_key, delta=delta)
```
ericholscher, line 317: Neat.

adserver/views.py
```diff
@@ -782,6 +782,12 @@ def handle_action(self, request, advertisement, offer, publisher):
                 request, self.impression_type, publisher=publisher, offer=offer
             )
 
+            # Update the publisher daily earn for cap calculations
+            if self.impression_type == CLICKS and advertisement.flight.cpc:
+                publisher.increment_daily_earn(float(advertisement.flight.cpc))
+            if self.impression_type == VIEWS and advertisement.flight.cpm:
+                publisher.increment_daily_earn(float(advertisement.flight.cpm) / 1000)
```
ericholscher, line 789: I do wonder if this will slowly drift given the float nature of it? Not a huge deal, but interesting.

ericholscher (approved):
This looks simple enough. I do worry that it's a bit simplistic, but seems like a good first approach to solve a problem, and we can always expand once we've learned more.

adserver/views.py
davidfischer: Floats should be accurate enough even to 4-5 decimals. I thought about it but I didn't think it would be an issue.

ericholscher: Yea, it's not billing data, so not important to get perfect.

davidfischer:
> This looks simple enough. I do worry that it's a bit simplistic, but seems like a good first approach to solve a problem, and we can always expand once we've learned more.

We may have to but I think this should work. I don't precisely know how atomic `cache.incr` is.

Eventually, we may end up storing everything needed to make a decision in the cache and having a batch that writes to the database but that's a much more significant rework.

PR #580, opened 2022-05-17
Allow publishers to create house ads

davidfischer:
This is still a work in progress but I wanted to show off what I'm working on here.

## In this PR

- Add a FK to Advertisers which makes them a "publisher specific" advertiser
- When creating a brand new publisher, it will also create an advertiser account (and campaign and a flight with a large number of impressions). We will have a data migration for existing publishers.
- Add a new campaign type called "publisher house" ads which are prioritized over regular house ads but below community ads.
- Adds the ability to target flights by domain or by publisher (include or exclude). This will be used for publisher house ads.
- Adds a screen in the publisher dashboard for "fallback ads" or something like that but it's basically just display logic that let's them upload ads to their flight. (**TODO**)

## Todo

There's still a few things left to do here:

- [ ] Publisher ad creation/update interface
- [ ] Data migration to create advertiser accounts for all our publishers


adserver/models.py
```diff
@@ -353,6 +354,16 @@ class Advertiser(TimeStampedModel, IndestructibleModel):
     name = models.CharField(_("Name"), max_length=200)
     slug = models.SlugField(_("Advertiser Slug"), max_length=200, unique=True)
 
+    # Publisher specific advertiser account
+    publisher = models.ForeignKey(
```
davidfischer, line 358: Perhaps a `OneToOneField` is better.

ericholscher: Seems likely. Though there might be some weird edge cases where multiple publishers want to run the same house ads? 🤷

ericholscher (approved):
Looks simple enough. I guess we can use the existing publisher reports to show the house ads for now, so they will at least get basic reporting that way 👍

PR #591, opened 2022-06-15
Add a textacy/spacy-based analyzer model

davidfischer:
- Uses the keywords from the database
- Notably handles phrases in text like "data science"
- Lemmatizes (canonicalizes) keywords and phrases
- Found keywords/phrases are ordered by importance/appearance

### DIFF
--- a/adserver/analyzer/backends/naive.py
+++ b/adserver/analyzer/backends/naive.py
@@ -3,7 +3,7 @@ import collections
 
 from bs4 import BeautifulSoup
 
-from ... import regiontopics as topics
+from ...models import Topic
 from .base import BaseAnalyzerBackend
 
 
@@ -31,17 +31,12 @@ class NaiveKeywordAnalyzerBackend(BaseAnalyzerBackend):
         """Overrides to get the keyword corpus."""
         super().__init__(url, **kwargs)
 
-        self.keyword_corpus = (
-            topics.data_science
-            + topics.backend_web
-            + topics.frontend_web
-            + topics.security_privacy
-            + topics.devops
-            + topics.python
-            + topics.game_dev
-            + topics.blockchain
-            + topics.techwriting
-        )
+        self.topics = Topic.load_from_cache()
+        self.keyword_corpus = []
+
+        for topic in self.topics:
+            for kw in self.topics[topic]:
+                self.keyword_corpus.append(kw)
 
     def analyze_response(self, resp):
         """Analyze an HTTP response and return a list of keywords/topics for the URL."""
@@ -55,11 +50,14 @@ class NaiveKeywordAnalyzerBackend(BaseAnalyzerBackend):
             # If no results, go to the next selector
             # If results are found, use these and stop looking at the selectors
             if results:
-                text = results[0].get_text().replace("\n", " ")
+                text = self.preprocess_text(results[0].get_text())
                 keywords = self.analyze_text(text)
                 break
 
         return keywords
+
+    def preprocess_text(self, text):
+        return text.replace("\n", " ")
 
     def analyze_text(self, text):
         """Analyze a large string of text for keyword extraction."""
--- a/adserver/analyzer/backends/textacynlp.py
+++ b/adserver/analyzer/backends/textacynlp.py
@@ -1,10 +1,14 @@
 """Textacy/Spacy-based keyword analyzer that lemmatizes and uses textrank for keyphrase extraction."""
 import collections
+import logging
 
 import textacy
 from textacy import preprocessing
 
 from .naive import NaiveKeywordAnalyzerBackend
+
+
+log = logging.getLogger(__name__)  # noqa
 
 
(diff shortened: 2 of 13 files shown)


adserver/analyzer/backends/naive.py
```diff
@@ -31,17 +31,12 @@ def __init__(self, url, **kwargs):
...
-            + topics.backend_web
-            + topics.frontend_web
-            + topics.security_privacy
-            + topics.devops
-            + topics.python
-            + topics.game_dev
-            + topics.blockchain
-            + topics.techwriting
-        )
+        self.topics = Topic.load_from_cache()
```
ericholscher, line 34: This workflow is way nicer ❤️

adserver/analyzer/backends/textacynlp.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def preprocess_text(self, text):
+        return self.preprocessor(text).lower()
+
+    def analyze_text(self, text):
+        """
+        Analyze a large string of text for keyword extraction.
+
+        One notable difference from the other backends is that the matching
+        keywords/phrases should be in order of how used/present they are.
```
ericholscher, line 64: 💯  comment

adserver/analyzer/backends/textacynlp.py
```diff
@@ -0,0 +1,95 @@
...
+        # Reset the keywords
+        self.keyword_corpus = []
+
+        # Build up the lemmatized keywords
+        for topic in self.topics:
+            for keyword in self.topics[topic]:
+                terms_doc = textacy.make_spacy_doc(
+                    self.preprocessor(keyword).lower(), lang=self.pretrained_model
+                )
+                words = list(
```
ericholscher, line 49: So we're basically pre-computing the lemma for every keyword here? Would be curious what this actually looks like, I'll plan to drop into a PDB locally and step through it.

adserver/analyzer/backends/textacynlp.py
```diff
@@ -0,0 +1,95 @@
...
+        One notable difference from the other backends is that the matching
+        keywords/phrases should be in order of how used/present they are.
+        """
+        doc = textacy.make_spacy_doc(text, lang=self.pretrained_model)
+        keyword_counter = collections.Counter()
+
+        # Lemmatize the words/phrases to make sure we're always finding exactly what we're looking for
+        # Bias toward the early part of the document
+        for phrase, weight in textacy.extract.keyterms.textrank(
+            doc, normalize="lemma", position_bias=True, topn=self.TOP_PHRASE_COUNT
```
ericholscher, line 72: Interesting that we aren't inputting our keywords here. It seems like that would help make more educated guesses around what we're wanting to pull out. Seems like there's definitely some interesting optimization possibilities here with other models, etc.

adserver/analyzer/backends/textacynlp.py
```diff
@@ -0,0 +1,95 @@
...
+                    # The keyphrase is present in the keyterms
+                    # However, make sure those appear on word boundaries
+                    # Or the word is long enough where that shouldn't matter
+                    # (eg. avoid the "ai" in "email" problem)
+                    keyphrase_set = set(keyphrase.split())
+                    if any(
+                        (
+                            len(keyphrase) > self.MIN_PHRASE_LENGTH,
+                            keyphrase_set
+                            == keyphrase_set.intersection(set(phrase.split())),
```
ericholscher, line 87: So this is basically checking for an exact match? 
```suggestion
                            keyphrase_set  # Check for exact match
                            == keyphrase_set.intersection(set(phrase.split())),
```

adserver/analyzer/backends/textacynlp.py
```diff
@@ -0,0 +1,95 @@
...
+                        )
+                    ):
+                        keyword_counter[keyphrase] += weight
+
+        return [
+            self.lemmatized_to_original_map[kw]
+            for kw, _ in keyword_counter.most_common(self.MAX_KEYWORDS)
```
ericholscher, line 94: Could bump up MAX_KEYWORDS above 3 once we're more comfortable.

davidfischer:
> we're still not using the DB keywords for live targeting, right?

Correct.

adserver/analyzer/backends/textacynlp.py
davidfischer: That's a weird place for that comment right? Before the double equals sign?

It ensures there's a word match (as the comment above suggests). If you look for a short word like "ai" that the word "ai" is there directly rather than just part of a larger word like "email". This creates a set of words and then checks that the all the words in our interested keyword set are present in the textrank phrase.

adserver/analyzer/backends/textacynlp.py
davidfischer: I haven't found something exactly like that in any library I looked at. Instead of analyzing some text and trying to reason about it, you're trying to classify how close a text is to some predefined topic set. I didn't find something that did that exactly.

PR #592, opened 2022-06-17
Fix the flight ordering for good

davidfischer:
Fixed in a way that https://github.com/readthedocs/ethical-ad-server/pull/466 didn't quite fix.

Based on flights being delayed or overlapping other flights (startdate1 before startdate2 but enddate1 after enddate2) you can have our flight headers get duplicated.

## Illustration of the problem
Screenshot from 2022-06-17 13-33-50

### DIFF
--- a/adserver/models.py
+++ b/adserver/models.py
@@ -811,7 +811,13 @@ class Flight(TimeStampedModel, IndestructibleModel):
 
     @classmethod
     def order_flights(cls, flights):
-        """Orders flights by our normal ordering of state, then date, then name."""
+        """
+        Orders flights by our normal ordering of state, then date, then name.
+
+        Previously, we were using database ordering, but that cannot deterministically
+        order flights by state. Flights can be upcoming if they start after today
+        or if they start before today (but end after today) but aren't live.
+        """
         state_order = {
             FLIGHT_STATE_CURRENT: 0,
             FLIGHT_STATE_UPCOMING: 1,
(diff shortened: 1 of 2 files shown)


adserver/models.py
```diff
@@ -809,6 +809,19 @@ def state(self):
...
+        """Orders flights by our normal ordering of state, then date, then name."""
+        state_order = {
+            FLIGHT_STATE_CURRENT: 0,
+            FLIGHT_STATE_UPCOMING: 1,
+            FLIGHT_STATE_PAST: 2,
+        }
+        return sorted(
+            flights,
+            key=lambda f: (state_order[f.state], -1 * f.start_date.toordinal(), f.name),
+        )
```
ericholscher, line 823: Would it be easier to use the `order_by` with multiple arguments? https://docs.djangoproject.com/en/4.0/ref/models/querysets/#django.db.models.query.QuerySet.order_by

In general it's nice to keep querysets as querysets, whereas this is turning them into a list. Probably fine for most use cases, but could be nicer if there isn't a reason not to use `order_by`.

davidfischer: Previously, we were using `order_by` with 3 arguments. However, that does not deterministically order the states (which is not a field). We would have to either store the state or change what the state means:

* A flight is currently upcoming if it is live or not as long as the start date is after today. This causes issues with DB ordering because ordering by start date or the live date doesn't definitely do it.
* A flight is current only if it is live AND the start date is before today
* Past is anything else

ericholscher: Makes sense. Perhaps worth including this in a comment so we remember why we did this in the future.

PR #596, opened 2022-06-21
Fix filtering of payouts that are paid

ericholscher:
This was written before we had a status field.
I'd love to have a nicer way to handle this,
but at least fixing the bug for now..


ericholscher:
Gonna merge this since it's staff-only. All this logic needs some reworking in the future, but it's at least useful to fix the bug for now.

davidfischer (approved):
I ignored the linting issue but there's a logic error either with this or with the test.

ericholscher:
Gonna close this. It's a silly implementation, and should be reworked at some point.

PR #598, opened 2022-06-23
Use analyzer keyword findings in ad decision

davidfischer:
**NOTE:** My goal isn't to merge this until we're ready to roll out the analyzer into wider use.

Uses keywords/topics found by the adserver analyzer during the decision process. This is merged with publisher keywords and ad client keywords.

## Considerations

- This will add an extra query to the ad decision process in most cases. That should add 3-5ms, I'd estimate.
- We could consider caching this but there's a few things to consider there. For very frequently used URLs, that would probably result in some speedup. However, if we use a local memory cache, it will only yield gains for very frequently used URLs that are hit by the same backend process (which are restarted every 10k requests) multiple times. However, if we add it to Redis to be shared by backend processes, then it might add about as much as the DB.

### DIFF
--- a/adserver/tests/test_decision_engine.py
+++ b/adserver/tests/test_decision_engine.py
@@ -7,6 +7,7 @@ from django.test.client import RequestFactory
 from django_dynamic_fixture import get
 from user_agents import parse
 
+from ..analyzer.models import AnalyzedUrl
 from ..constants import AFFILIATE_CAMPAIGN
 from ..constants import CLICKS
 from ..constants import COMMUNITY_CAMPAIGN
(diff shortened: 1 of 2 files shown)


adserver/tests/test_decision_engine.py
```diff
@@ -613,9 +614,34 @@ def test_default_keywords(self):
             request=self.request, placements=self.placements, publisher=self.publisher
         )
 
-        self.assertTrue(
-            self.backend.keywords, ["foo", "bar", "baz", "machine-learning"]
+        self.assertEqual(
```
davidfischer, line 617: While running the tests, I realized that this was just plain wrong. It was working but the test wasn't correctly checking it.

ericholscher (commented):
This looks quite simple. Definitely some considerations before we ship it, but I don't worry too much about performance currently.

PR #605, opened 2022-07-05
Fix multi topic targeting bug

davidfischer:
Without this fix, flights targeting multiple topics will get zero views.

### DIFF
--- a/adserver/models.py
+++ b/adserver/models.py
@@ -914,7 +914,11 @@ class Flight(TimeStampedModel, IndestructibleModel):
         # Check topics (groupings of keywords)
         if self.included_topics:
             topics = Topic.load_from_cache()
-            topic_keyword_set = set(*[topics[slug] for slug in self.included_topics])
+            topic_keyword_set = set()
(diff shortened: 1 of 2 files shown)


adserver/models.py
```diff
@@ -914,7 +914,11 @@ def show_to_keywords(self, keywords):
         # Check topics (groupings of keywords)
         if self.included_topics:
             topics = Topic.load_from_cache()
-            topic_keyword_set = set(*[topics[slug] for slug in self.included_topics])
+            topic_keyword_set = set()
+            for topic_slug in self.included_topics:
+                for kw in topics[topic_slug]:
+                    topic_keyword_set.add(kw)
```
ericholscher, line 920: This further validates my mistrust of highly functional code 😀

ericholscher (approved):

adserver/models.py
davidfischer: Can't argue with you there. I think I could have fixed it with a compound comprehension. However, trying to be clever is how I created this mess.

PR #609, opened 2022-07-14
Add the topic classifier backend

davidfischer:
This is a simple backend for the classifier. It outputs the topic categories as the keywords.

## Considerations

* Possibly we should do keyword extraction as well (maybe using the default textacy model?) and use those keywords as well. This will be useful for advertisers not targeting our major topics but rather something long tail (eg. `two-factor-auth`).
* The model threshold (currently 0.4) is a debatable number. I feel 0.4-0.5 are reasonable.


ericholscher (approved):
Simple enough! ❤️ 

I agree .4 is a reasonable place to start, and see how things turn out!

PR #610, opened 2022-07-14
Pass keywords to the ad rendering code

ericholscher:
This allows our ad templates to take into account the keywords on the request,
so we can be smarter about linking to topic-specific landing pages.

### DIFF
--- a/adserver/models.py
+++ b/adserver/models.py
@@ -1533,6 +1533,14 @@ class Advertisement(TimeStampedModel, IndestructibleModel):
         text = self.render_links(click_url)
         body = html.unescape(bleach.clean(text, tags=[], strip=True))
 
+        # Match keywords to topics
+        topics = Topic.load_from_cache()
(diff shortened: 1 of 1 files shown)


davidfischer (approved):
Seems simple enough. While I think this could get complicated since we don't want to be doing word stemming or other complex stuff in the template, I think it might enable some cool things in the simple cases.

ericholscher:
@davidfischer Thinking more on this, I wonder if we want to be smart in the backend here, and just pass a set of topics? That seems like what we actually want.

davidfischer:
> hinking more on this, I wonder if we want to be smart in the backend here, and just pass a set of topics? That seems like what we actually want.

I think that's better, TBH.

ericholscher:
@davidfischer Review request here :)

davidfischer (approved):
Looks easy enough.

PR #613, opened 2022-07-19
Add a screen for authorized users for a publisher

davidfischer:
Uses the identical form as for advertisers. We've had 2 requests for this in the last couple weeks so seemed worth it.

### DIFF
--- a/adserver/views.py
+++ b/adserver/views.py
@@ -2261,6 +2261,107 @@ class StaffPublisherReportView(BaseReportView):
         return context
 
 
+class PublisherAuthorizedUsersView(PublisherAccessMixin, UserPassesTestMixin, ListView):
+
+    """Authorized users for a publisher."""
+
+    context_object_name = "users"
+    model = get_user_model()
+    template_name = "adserver/publisher/users.html"
+
+    def get_context_data(self, **kwargs):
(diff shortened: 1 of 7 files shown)


adserver/views.py
```diff
@@ -2261,6 +2261,107 @@ def get_context_data(self, **kwargs):  # pylint: disable=too-many-locals
...
+
+
+class PublisherAuthorizedUsersRemoveView(
+    PublisherAccessMixin, UserPassesTestMixin, TemplateView
+):
+
+    """
+    Remove authorized users for a publisher.
+
+    This doesn't remove or deactivate the user - just removes them from the publisher.
```
ericholscher, line 2328: If they login, they will just be shown a screen with no publishers/advertisers right?

ericholscher (approved):
Definitely useful for a subset of publishers 👍

adserver/views.py
davidfischer: That is correct. They get a a message "You don't have access to anything": https://github.com/readthedocs/ethical-ad-server/blob/5f7dc3b30510f80266f99bf261abbbfbe85b2348/adserver/templates/adserver/dashboard.html#L58

PR #615, opened 2022-07-21
Fix up the model for release

davidfischer:
- Remove game-dev, techwriting, and hardware topics (for now)
- Remove navigation from the analyzer
- Add additional examples that weren't classified well

### DIFF
--- a/adserver/analyzer/backends/naive.py
+++ b/adserver/analyzer/backends/naive.py
@@ -23,6 +23,18 @@ class NaiveKeywordAnalyzerBackend(BaseAnalyzerBackend):
         "body",
     )
 
+    # CSS selectors of content that should be ignored for the purpose of analysis
+    REMOVE_CONTENT_SELECTORS = (
+        "[role=navigation]",
+        "[role=search]",
+        ".headerlink",
+        "nav",
+        "footer",
+        "div.header",
+        # Django Packages specific
+        "#myrotatingnav",
+    )
+
     MAX_WORDS_ANALYZED = 1000
     MAX_KEYWORDS = 3
     MIN_KEYWORD_OCCURRENCES = 2
@@ -43,6 +55,10 @@ class NaiveKeywordAnalyzerBackend(BaseAnalyzerBackend):
         keywords = []
 
         soup = BeautifulSoup(resp.content, features="html.parser")
+
+        for selector in self.REMOVE_CONTENT_SELECTORS:
+            for nodes in soup.select(selector):
+                nodes.decompose()
 
         for selector in self.MAIN_CONTENT_SELECTORS:
             results = soup.select(selector, limit=1)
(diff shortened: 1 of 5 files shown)


davidfischer:
```
================================== Results ==================================

TOK                   100.00
TEXTCAT (macro AUC)   90.32 
SPEED                 28336 


=========================== Textcat F (per label) ===========================

                   P       R       F
datascience    90.91   90.91   90.91
backend        75.00   18.75   30.00
frontend       50.00   11.11   18.18
security       66.67   18.18   28.57
devops         81.25   56.52   66.67
blockchain    100.00   80.00   88.89


======================== Textcat ROC AUC (per label) ========================

              ROC AUC
datascience      0.96
backend          0.79
frontend         0.87
security         0.91
devops           0.89
blockchain       0.99
```
Still more to go but getting better.

adserver/analyzer/backends/naive.py
```diff
@@ -23,6 +23,18 @@ class NaiveKeywordAnalyzerBackend(BaseAnalyzerBackend):
         "body",
     )
 
+    # CSS selectors of content that should be ignored for the purpose of analysis
```
ericholscher, line 26: This is the same logic that was in the training import code, just ported to the backend, right?

machine_learning_experiments/scripts/preprocess.py
```diff
@@ -13,10 +13,10 @@
     "frontend",
     "security",
     "devops",
-    "game-dev",
+    # "game-dev",
```
ericholscher, line 16: I'm a little confused on our semantics around slugging here. I would expect `data-science` and `tech-writing` to be hyphenated if we're doing that, but it seems we're not consistent. Should we define how we're slugging things, so we can try to be consistent, as well as asking our users to be consistent.

ericholscher (approved):
Solid improvements. I'm a little worried about all the content we're tagging as "no topic" and how that might impact the model. I feel like there's tons of python/node stuff in the documentation tooling that will probably be focusing for it when it's also very present in backend tags for example. But perhaps that's how we make it more discerning. 

> frontend       50.00   11.11   18.18

I'm surprised frontend is so low. That definitely feels like the next area to improve. Perhaps because we don't have as much content-heavy sites in that area?

adserver/analyzer/backends/naive.py
davidfischer: Correct, ported to the analyzer.

machine_learning_experiments/scripts/preprocess.py
davidfischer: Sure. The important thing is that there has to be a keyword with a matching slug (we have a keyword "data science" and "datascience" so either way works).

davidfischer:
> I'm surprised frontend is so low. That definitely feels like the next area to improve. Perhaps because we don't have as much content-heavy sites in that area?

This is definitely an area for improvement.

PR #617, opened 2022-07-21
Limit the model input to 100k characters

davidfischer:
According to an [error message in Sentry](https://sentry.io/organizations/read-the-docs/issues/3437531888/?project=1846644&query=is%3Aunresolved)m they say that it uses ~1GB of memory per 100k characters. There's a hard cap at 1M characters.


ericholscher (approved):
Seems like a good default 👍

PR #619, opened 2022-07-29
Changes the name for new publisher house ads accounts

davidfischer:
- It doesn't change past accounts
- Makes these a bit more obvious and easy to find


agjohnson (approved):

ericholscher (commented):
This seems reasonable 👍

PR #620, opened 2022-07-29
Move advertiser overview mostly to metabase

davidfischer:
This moves most of the advertiser overview display to metabase. This has some advantages and disadvantages. The PR doesn't change what data is displayed -- only how it is displayed.

## Advantages

* Metabase automatically handles long running queries, caches them, and uses loading indicators
* This should completely resolve timeouts on this page. All queries should be *very* fast.
* We can actually modify the dashboard (say add a new question or change one) and change production without deploying new code

## Disadvantages

* Loading data out of an iframe (metabase uses iframes) results in some weird positioning especially on mobile. There were zero mobile visits by advertisers to the dashboard in the last month. I attached a screenshot of this.

## Screenshots
Screenshot from 2022-07-29 14-46-38
Screenshot from 2022-07-29 14-52-01


agjohnson (approved):
Looks great! I think this might have been the change that we discussed in chat already. Without mobile views on the dashboard, this seems completely fine. If we get mobile views, we could always hide the iframe with CSS too.

ericholscher:
This looks great. 👍

PR #621, opened 2022-07-29
Offload the advertiser geo report to metabase

davidfischer:
- Still reports on a geo (not region) basis
- Works for basically any time frame even big ones on huge advertisers (no timeouts)
- I don't love the ranges metabase chooses but it doesn't look customizable. You can customize the colors but not set thresholds

## Testing

- Set `METABASE_SITE_URL=https://ethicalads.metabaseapp.com` and `METABASE_SECRET_KEY` in `.envs/local/django`
- Ensure an advertiser slug locally matches a production advertiser slug
- Load the geo report

## Screenshot
Screenshot from 2022-07-29 15-39-28


adserver/templates/adserver/reports/advertiser-geo.html
```diff
@@ -19,42 +17,22 @@
...
-  <h3>About this report</h3>
-  <p>
-    This report shows all your traffic that had country data we could determine from the IP address.
-  </p>
-  <em>
-  This report shows the <strong>top {{ limit }} geos</strong> and updates periodically. All previous days data is complete.
-  </em>
+  <h3>{% trans 'About this report' %}</h3>
+  <p>{% trans 'This report shows all your traffic that had country data we could determine from the IP address.' %}</p>
+  <em>{% trans 'All previous days data is complete. This report can take up to a minute to load.' %}</em>
```
agjohnson, line 24: I'm not familiar with whether this is a singular previous day or plural previous days, but I'm assuming plural possessive is wanted here:

```suggestion
  <em>{% trans 'All previous days' data is complete. This report can take up to a minute to load.' %}</em>
```

agjohnson (commented):
We talked about this one a little in chat, and there are some minor concerns with the metabase dashboard injection. But, the benefits outweight these it seems, the dashboard view is quite a nice improvement.

I don't have full history on this one, so we could wait to merge on this too, but technically speaking this change looks good

adserver/templates/adserver/reports/advertiser-geo.html
davidfischer: I can't take this suggestion directly as it causes a template syntax error with the translation. We could just do "All data for previous days is complete".

davidfischer:
> We talked about this one a little in chat, and there are some minor concerns with the metabase dashboard injection. But, the benefits outweight these it seems, the dashboard view is quite a nice improvement.

On this report, I'm less concerned about metabase iframes being mobile unfriendly than I am on the advertiser overview screen (#620). We've deployed a couple more detailed reports that are metabase exclusive.

adserver/templates/adserver/reports/advertiser-geo.html
agjohnson: Oops, missed the string quote style, yeah. I like your suggestion more either way.

agjohnson (approved):

PR #623, opened 2022-08-01
Log some client mismatches to the security logger

davidfischer:
After looking at the data, we can decide to filter these completely.


agjohnson (approved):
:+1:

The IP comparison will yield some false positives, as there are some legitimate cases for this scenario. But, either way, all of these are great to log and track.

davidfischer:
> The IP comparison will yield some false positives, as there are some legitimate cases for this scenario. But, either way, all of these are great to log and track.

I expect it will. I'm interested in those scenarios and how frequent they are.

PR #629, opened 2022-08-16
Ignore mismatched browsers/OSs

davidfischer:


adserver/views.py
```diff
@@ -753,28 +753,30 @@ def ignore_tracking_reason(self, request, advertisement, offer):
...
             log.log(
                 self.log_security_level,
                 "Mismatched browser between offer and impression. Publisher: [%s], Offer Browser: [%s], User agent: [%s]",
                 offer.publisher,
                 offer.browser_family,
                 user_agent,
             )
+            reason = "Mismatched browser"
+
+        # This is out of the elif block and will be run everytime
```
ericholscher, line 775: Do we actually care about this happening? It seems to happen enough that I wonder if logging it is even useful -- I guess perhaps for debugging or fraud detection, but I don't know if we'll really look at it.

ericholscher (approved):

adserver/views.py
davidfischer: By lowering the level, it won't make it into Sentry but I think this can signal that something is wrong for a publisher.

PR #634, opened 2022-08-31
Note about campaigns running over

davidfischer:
We have a few campaigns running long right now. I just wanted to add a small note about it. It isn't super visible but might head off some support questions.


ericholscher (approved):
Pretty minimal, but useful 👍

PR #637, opened 2022-09-16
Add a GPU config for the model

davidfischer:
When running on my laptop, I get:

```
UserWarning: User provided device_type of 'cuda', but CUDA is not available. Disabling
```

Also, it is MUCH MUCH slower than normal training. I'm interested to run this on a GPU in the cloud.


adserver/analyzer/tests.py
```diff
@@ -174,8 +174,8 @@ def test_analyzer_success(self):
 
                 <!-- Only this should be used -->
                 <main>
-                <p>Data science and machine learning are important</p>
```
davidfischer, line 177: Something changed with the lemmatizer and using the built-in model. It is lemmatizing differently than before. This does *not* affect our trained model but only the built-in model.

ericholscher (approved):
Looks pretty simple. I think on the call they mentioned using the transformer model as the base for this -- are we still using the `en_core_web_md` model?

davidfischer:
> Looks pretty simple. I think on the call they mentioned using the transformer model as the base for this -- are we still using the `en_core_web_md` model?

This isn't used with our trained model but is used in the standard textacy model (which is used in tests since ours isn't installed).

PR #638, opened 2022-09-23
Add a management command for ease of running the model in dev

davidfischer:
You can run the model easily against any URL with our exact preprocessing with the following:

```bash
ADSERVER_ANALYZER_BACKEND=adserver.analyzer.backends.EthicalAdsTopicsBackend ./manage.py runmodel https://example.com
```

Results are printed to the terminal but not stored anywhere. This is for development primarily but could be run in prod without affecting prod.


ericholscher (approved):
Looks useful! Worth putting in some docs somewhere?

ericholscher:
👍

PR #639, opened 2022-09-26
Do language detection in the model

davidfischer:
Ignore non-English for now.


ericholscher (approved):
Seems simple enough. I do wonder about logging above debug for this in production to start, so we know how much this is happening? I'd guess a lot, since we're running it over a bunch of RTD sites that aren't in English? Might be too noisy then..

davidfischer:
It's actually not too hard to see how much this is happening at least broadly. Look for RTD pages that aren't `/en/`.

PR #640, opened 2022-09-26
Add more categorized data for the model

davidfischer:
This adds more categorized data for the model. We still need more but probably best to start on frontend and backend. Having 100+ of each should be good (currently there's ~50-60)

### DIFF
--- a/machine_learning_experiments/assets/categorized-data.yml
+++ b/machine_learning_experiments/assets/categorized-data.yml
@@ -269,6 +269,10 @@
   topics: ["datascience"]
 - url: "https://docs.deepmodeling.com/projects/deepmd/en/master/getting-started/training.html"
   topics: ["datascience"]
+- url: "https://mmcv.readthedocs.io/en/latest/get_started/introduction.html"
+  topics: ["datascience"]
+- url: "https://mmcv.readthedocs.io/en/latest/get_started/previous_versions.html"
+  topics: ["datascience"]
 
 # Conda official docs
 - url: "https://docs.conda.io/en/latest/"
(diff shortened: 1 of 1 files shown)


machine_learning_experiments/assets/categorized-data.yml
```diff
@@ -948,3 +1027,31 @@
...
+  topics: []
+- url: "https://origins.readthedocs.io/en/latest/types/damage_condition_types/attacker/"
+  topics: []  # Game dev
+- url: "https://androidaps.readthedocs.io/en/latest/Usage/Enabling-Engineering-Mode-in-xDrip.html"
+  topics: []  # Maybe mobile
+- url: "https://hikari-lightbulb.readthedocs.io/en/latest/index.html"
+  topics: []
+- url: "https://hikari-lightbulb.readthedocs.io/en/latest/guides/application-commands.html"
+  topics: []
+- url: "https://google-calendar-simple-api.readthedocs.io/en/latest/events.html"
```
ericholscher, line 1048: I'm a little hesitant to tag pages like this as no topic, since they have a ton of Python code on them. It definitely feels like this would be very confusing to me if you told me this wasn't backend, but then told me other pages with random Python code were.

ericholscher (commented):
This seems good, but I do think untagging a bunch of Python stuff is confusing. I'm not convinced that this will get us better results, and perhaps degrade how well out backend tagging works?

machine_learning_experiments/assets/categorized-data.yml
davidfischer: I think this requires more discussion. Certainly the argument could be made that most Python stuff is "backend" although it isn't exactly web.

ericholscher: Yea.. I think it's probably just a symptom of having non-exclusive topics. There's always going to be something that matches 80% nothing and 20% an existing topic, or similar. 🤷 

I think it's fine to move ahead here, since this is a specific instance of a larger problem, and we'll just continue to refine to get closer to a solution over time.

ericholscher (approved):

davidfischer:
I'm going to move this over to the new repo (https://github.com/readthedocs/ethicalads-model) and close this PR.

davidfischer:
The ML model moved to it's own repo. Closing.

PR #642, opened 2022-09-27
Report will use our regions and topics from the DB

davidfischer:
The keyword report has been giving incorrect results for flights targeting by topic.

The regiontopic report will use the set of non-overlapping regions so we can get a single region for a country.


adserver/regiontopics.py
```diff
@@ -1,320 +0,0 @@
-"""Our targeting data for topics & regions."""
```
ericholscher, line 1: I did actually find this useful for linking to users to show what keywords are default in each region, as well as copy/pasting the list of country codes for doing country-based targeting. It's definitely not the best solution for either of those, but I do think having some way to link someone this information is useful. Perhaps an API that just returns this data for each of our regions/topics?

adserver/templates/adserver/reports/staff-regiontopics.html
```diff
@@ -21,17 +21,17 @@
...
     {% endfor %}
   </select>
 </div>
 <div class="col-xl-3 col-md-6 col-12 mb-3">
   <label class="col-form-label" for="id_topic">{% trans 'Topic' %}</label>
   <select class="form-control" name="topic" id="id_topic">
     <option value="">All Topics</option>
-    {% for name in topic_list %}
-      <option value="{{ name }}"{% if topic == name %} selected{% endif %}>{{ name }}</option>
+    {% for tp in topic_list %}
```
ericholscher, line 33: It's probably nicer to use `region` and `topic` here, unless we're worried about it clobbering that variable from somewhere else?

adserver/regiontopics.py
davidfischer: An API seems reasonable although that's still a little challenging to "show" customers. Maybe documentation as well?

adserver/templates/adserver/reports/staff-regiontopics.html
davidfischer: That is exactly what I'm avoiding. `region` and `topic` are set in the context.

adserver/regiontopics.py
ericholscher: Yea, it would be best if the docs could render via API, or at least build from the API, since the topics change over time.

ericholscher:
@davidfischer This looks great though, I think we should ship it! Maybe do a quick deploy today, so I can reindex the ClearML campaigns?

ericholscher:
I ended up just checking this branch out on ads-extra, and running the index in a `tmux`.

PR #647, opened 2022-10-14
Add a custom cache time for publishers

davidfischer:
We usually cache ads for a certain amount of time on each publisher. This change lets us have a custom cache time specific to a publisher.


adserver/models.py
```diff
@@ -393,6 +393,13 @@ class Publisher(TimeStampedModel, IndestructibleModel):
             "Cache this publishers ad requests. Disable for special cases (eg. SaaS users)"
         ),
     )
+    cache_ads_duration = models.PositiveIntegerField(
+        default=0,
```
ericholscher, line 397: 0 seems a bit confusing for a default here, I would expect it to mean no caching or infinite caching. Do we want to use None instead?

ericholscher (approved):

adserver/models.py
davidfischer: I try not to have nullable integer fields but if you feel strongly, I'll change it.

ericholscher: I don't feel too strongly. Maybe just note it in the docstring.

PR #656, opened 2022-11-01
Skip recently analyzed URLs

davidfischer:


davidfischer:
CI issues fixed in https://github.com/readthedocs/ethical-ad-server/pull/657

PR #657, opened 2022-11-01
Shuffle dependencies

davidfischer:
Specifically, move pytorch (a spacy-transformer dep) into the production file. This avoids a 250MB install in CI.


ericholscher (approved):
Makes sense 👍

PR #662, opened 2022-11-10
Split the dashboard view so it scales a bit better with more data

ericholscher:
This makes it easier to get to the advertisers quickly in a staff view. 

Screenshot 2022-11-10 at 2 18 55 PM

### DIFF
--- a/adserver/models.py
+++ b/adserver/models.py
@@ -1772,6 +1772,8 @@ class Advertisement(TimeStampedModel, IndestructibleModel):
             # Don't do this by default as searching for a template is expensive
             template = get_template("adserver/advertisement.html")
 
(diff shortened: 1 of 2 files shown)


adserver/models.py
```diff
@@ -1772,6 +1772,8 @@ def render_ad(
             # Don't do this by default as searching for a template is expensive
             template = get_template("adserver/advertisement.html")
 
+        log.debug(f"Rendering Ad. topics={topics}")
```
ericholscher, line 1775: Woops, accidental commit

davidfischer (approved):
Sounds good. I was thinking we should have a searchable field but this is probably just as good for now.

PR #663, opened 2022-11-18
Optimize the keyword aggregation

davidfischer:
In my testing, this should cut this aggregation down to ~15m. This follows a pattern similar to what we're doing for regions and topics.


ericholscher (approved):
Simple enough 👍

PR #664, opened 2022-11-18
Move publisher reports to metabase

davidfischer:
Some of these reports are timing out for large publishers and offloading to metabase will allow them to load async. I verified that the numbers in our tables match the reports.

## Screenshots
Screenshot from 2022-11-18 13-40-22
Screenshot from 2022-11-18 13-51-45


ericholscher (approved):
Nice 💯

PR #672, opened 2022-12-09
Pass the `topics` to the template string for advertisers

ericholscher:
This lets them capture that information based on the keywords in the offer.
This could pass multiple topics for an ad that is only targeting one,
so I'm not sure if we want to match this against the campaign or not,
but that would add additional processing.


ericholscher:
Nobody has asked for this, so closing for now.

PR #674, opened 2022-12-19
Add callout on flight detail to contact us for changes.

ericholscher:
I get a few people asking about this each month,
so would be good to be explicit.


davidfischer (approved):
Great change!

PR #676, opened 2023-01-03
Staff publisher form handles an existing user

davidfischer:


ericholscher (approved):
Is this in response to a bug or a user hitting this?

davidfischer:
Yes, a bug I triggered. We already do this for advertisers. We had a publisher want a second account after selling their first site to somebody else.

PR #677, opened 2023-01-05
Use GitHub Actions for CI

davidfischer:
Currently this takes ~7m to run since it isn't run in parallel (CircleCI ran the lint checks in parallel with the unit tests). We could run them in parallel or perhaps better yet we could just have a new tox target that installs the dependencies once and then runs all the commands. It would probably save a couple minutes on checks.


ericholscher (approved):
This seems like a good start. I agree it probably makes sense to make a fancier tox setup that shares envs, but this is fine to get started 👍

PR #680, opened 2023-01-12
Speed up the ads and campaign admin views

davidfischer:
Basically just shows less data. Users can dig into individual objects if they want this data.


ericholscher (approved):

davidfischer:
I'll wait for the CI fix and then merge it. There's a new version of an indirect dep messing things up.

ericholscher:
@davidfischer Might be worth using pip-tools similar to RTD, which pins all implicit deps.

davidfischer:
I tend to agree that having a similar setup to RTD would be good. Also pinning all deps is worthwhile.

PR #681, opened 2023-01-13
Add optimized publisher and advertiser indexes

davidfischer:
Indexes publisher per day and advertiser per day. Since the aggregation tasks use the `AdImpression` table instead of `Offer`, they are much faster. A day's aggregation can be run in a few seconds.


adserver/models.py
```diff
@@ -1883,6 +1883,68 @@ def __str__(self):
         return "%s on %s" % (self.advertisement, self.date)
 
 
+class AdvertiserImpression(BaseImpression):
```
ericholscher, line 1886: I'd probably name this a bit more descriptively? `spend` feels pretty jargony, is there a more classic name for this? I can't think of one offhand tho.. `cost` or `expense`, but don't love either. 

```suggestion
class AdvertiserSpendImpression(BaseImpression):
```

adserver/models.py
```diff
@@ -1883,6 +1883,68 @@ def __str__(self):
...
+        ordering = ("-date",)
+        unique_together = ("advertiser", "date")
+        verbose_name_plural = _("Advertiser impressions")
+
+    def __str__(self):
+        """Simple override."""
+        return "%s on %s" % (self.advertiser, self.date)
+
+
+class PublisherImpression(BaseImpression):
```
ericholscher, line 1914:
```suggestion
class PublisherRevenueImpression(BaseImpression):
```

adserver/tasks.py
```diff
@@ -490,6 +493,106 @@ def daily_update_uplift(day=None):
...
+            "advertisement__flight__campaign__advertiser__name",
+            "advertisement__flight__campaign__advertiser_id",
+        )
+        .annotate(
+            total_decisions=Sum("decisions"),
+            total_offers=Sum("offers"),
+            total_views=Sum("views"),
+            total_clicks=Sum("clicks"),
+            total_spend=Sum(
+                (F("clicks") * F("advertisement__flight__cpc"))
```
ericholscher, line 525: I wish we had a nicer way to single source this stuff. It feels error prone to repeat it over & over, but it does tend to change slightly each time 🙃

ericholscher (approved):
Looks very useful!

adserver/tasks.py
davidfischer: I think the only way to really do that is stored procedures but I think they'll differ a bit each time.

adserver/models.py
davidfischer: It isn't really a "spend" or "cost" impression though. While that's one field, the other inherited fields (views, clicks, offers, decisions) are also what this aggregation is about. It really is an aggregation about an advertiser since it is just advertiser details on a daily basis. I don't think this is an improvement.

ericholscher: Hrm, I guess that's true. Just feels quite generic, but I guess I'm alright with it for now?

PR #683, opened 2023-01-17
Change analyzer threshold

davidfischer:
Scan a site no more than monthly.
Wait for 100 ad views between scans.


ericholscher (approved):
Easy enough :)

PR #688, opened 2023-01-19
Fix support link in flight metadata

davidfischer:
This wasn't always set depending on the page.
It was not set on the default flight metadata view.


ericholscher (approved):
Bleh, thanks. Could be worth including in our default context processors at some point 🤔

PR #690, opened 2023-01-19
Put the advertiser overview dashboard in the report screen

davidfischer:
- Replace the flight list. If users want a flight report, they can get it from the flight detail view.
- This allows users to effectively filter this dashboard by date.

### DIFF
--- a/adserver/templates/adserver/advertiser/flight-list.html
+++ b/adserver/templates/adserver/advertiser/flight-list.html
@@ -58,9 +58,22 @@
                 <td class="text-right">${{ flight.projected_total_value|floatformat:2 }}</td>
                 <td class="text-right">{{ flight.ctr|floatformat:3 }}%</td>
                 <td class="text-right">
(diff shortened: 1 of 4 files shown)


ericholscher (approved):
😍 

> Replace the flight list. If users want a flight report, they can get it from the flight detail view.

I wonder about adding a small 📈 icon to the flight listing in the `manage advertising` view, so that it's easier to jump to that flight report there as well?

ericholscher:
My idea is probably for another PR, not this one. But it always does throw me off that these 2 list views are so similar, but not the same:

Screenshot 2023-01-19 at 3 32 22 PM
Screenshot 2023-01-19 at 3 32 26 PM

davidfischer:
> My idea is probably for another PR, not this one. But it always does throw me off that these 2 list views are so similar, but not the same:
> 

Having them be the same seems fine.

davidfischer:
> I wonder about adding a small chart_with_upwards_trend icon to the flight listing in the `manage advertising` view, so that it's easier to jump to that flight report there as well?

Done.

ericholscher (approved):
🤩

PR #691, opened 2023-01-19
Run quick indexes periodically

davidfischer:
- Pretty frequent in dev and half hourly in prod
- Fixes a minor bug with celerybeat locally

### DIFF
--- a/config/settings/development.py
+++ b/config/settings/development.py
@@ -35,6 +35,7 @@ LOGGING["loggers"]["adserver"]["level"] = "DEBUG"
 CELERY_TASK_ALWAYS_EAGER = False
 CELERY_BROKER_URL = env("CELERY_BROKER_URL", default=env("REDIS_URL", default=None))
 CELERY_RESULT_BACKEND = CELERY_BROKER_URL
+CELERY_CREATE_MISSING_QUEUES = True
 
 CELERY_BEAT_SCHEDULE = {
     "dev-geo-index": {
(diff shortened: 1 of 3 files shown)


config/settings/development.py
```diff
@@ -35,6 +35,7 @@
 CELERY_TASK_ALWAYS_EAGER = False
 CELERY_BROKER_URL = env("CELERY_BROKER_URL", default=env("REDIS_URL", default=None))
 CELERY_RESULT_BACKEND = CELERY_BROKER_URL
+CELERY_CREATE_MISSING_QUEUES = True
```
davidfischer, line 38: I don't think this is strictly necessary but explicit seems good.

docker-compose/local/django/celery/beat/start
```diff
@@ -5,4 +5,4 @@ set -o nounset
 
 
 rm -f './celerybeat.pid'
-celery -A config.celery_app beat -l INFO -Q celery,analyzer
+celery -A config.celery_app beat -l INFO
```
davidfischer, line 8: This is not necessary on the beat call, only the worker call. We aren't doing this in prod and it was preventing beat from running in dev.

config/settings/development.py
ericholscher: Yea, I think it defaults on, but good to be exlicit.

docker-compose/local/django/celery/beat/start
ericholscher: Odd. I thought it would only listen for 1 by default.

ericholscher (approved):

PR #694, opened 2023-01-26
Add a metabase publisher dashboard

davidfischer:
- Uses our optimized indexes (should be fast)
- Focuses on CTR
- Shows view rate in details

## Testing
To test this against production data:

- Set `METABASE_SECRET_KEY` and `METABASE_SITE_URL` in `.envs/django`
- Create a publisher locally with a slug matching a production publisher
- View the dashboard. Not that figures will be off by a single day because this dashboard is running against our new optimized indexes which are only run daily in prod. After the next deploy they'll be run more often.

## Screenshot
Screenshot from 2023-01-25 16-57-15


adserver/views.py
```diff
@@ -221,16 +223,14 @@ def get_context_data(self, **kwargs):
...
+            advertiser_id=self.advertiser.id
         ).exists()
 
     def has_views_this_month(self, start_date):
         """Detect if advertisers have impressions since the start date (first of the month)."""
         return (
-            AdImpression.objects.filter(
-                advertisement__flight__campaign__advertiser_id=self.advertiser.id
-            )
+            AdvertiserImpression.objects.filter(advertiser_id=self.advertiser.id)
```
ericholscher, line 233: Seems this is snuck in with a Publisher PR, but good to get them both done at once :)

ericholscher (approved):
Nice! 💯

PR #696, opened 2023-02-01
Fix a template inheritance issue with advertiser reports

davidfischer:
We added a dashboard to the advertiser reports page. However, that template is inherited by some others causing the dashboard (which didn't work due to data) to be added on other pages. This will only show the dashboard if the data is present.


ericholscher (approved):
I always learn to be more defensive in code, every time something like this happens. Hard to know when it's overkill tho.

PR #700, opened 2023-02-07
Automate sending flight wrapup emails

davidfischer:
This will send them directly rather than leaving them as drafts.

Currently, our production Front Backend settings use @ericholscher's account with the "name" set to `EthicalAds by Read the Docs`. These can be customized in our private -ops repo. We could also set a custom name *just* for flight wrapups if we want to (eg. "EthicalAds Flight Tracker") with a couple modifications.

### DIFF
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -737,10 +737,6 @@ def notify_of_completed_flights():
                     to=to_addresses,
                     connection=connection,
                 )
-
-                # Make this a draft instead of just sending it directly
-                message.draft = True
-
                 message.send()
 
 
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -729,7 +729,10 @@ def notify_of_completed_flights():
                 ],
             }
 
-            with mail.get_connection(settings.FRONT_BACKEND) as connection:
+            with mail.get_connection(
+                settings.FRONT_BACKEND,
+                sender_name=f"{site.name} Flight Tracker",
+            ) as connection:
                 message = mail.EmailMessage(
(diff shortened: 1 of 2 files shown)


ericholscher (commented):
I feel like we want to modify the email content here, along with automating it.

Also I _love_ `EthicalAds Flight Tracker` as a name :)

davidfischer:
> I feel like we want to modify the email content here, along with automating it.

What content are you thinking of changing?

ericholscher:
Looking at it again, I think it's probably fine... I thought I wanted changes, but didn't have anything obvious except small tweaks. I do feel like the language avoiding pluralization is still a bit awkward, but not sure how to make it better.

adserver/tasks.py
```diff
@@ -729,7 +729,10 @@ def notify_of_completed_flights():
                 ],
             }
 
-            with mail.get_connection(settings.FRONT_BACKEND) as connection:
+            with mail.get_connection(
+                settings.FRONT_BACKEND,
+                sender_name=f"{site.name} Flight Tracker",
```
ericholscher, line 734: 😍

ericholscher (approved):

davidfischer:
I switched it to use "EthicalAds flight tracker" as the sender name. The sender name is *not* used if you send a draft, however, since once somebody clicks send they take over as the sender.

adserver/tasks.py
davidfischer: Screenshot from 2023-02-07 15-45-43

ericholscher:
Yep, works the same with all our EA automation (eg. publisher emails)

PR #702, opened 2023-02-10
Pace ads by a custom interval

davidfischer:
Currently, we pace ads a day at a time. So if a campaign is 1000 clicks over 10 days, we will aim to space it out so the campaign gets 100 clicks per day. Switching to a smaller interval (say, an hour) will have even more benefits like pacing the campaign more evenly across regions (not missing any because that region is asleep when the day resets).


davidfischer:
Considerations:

- Should we rename methods like `clicks_needed_today` to `clicks_needed_this_interval`? Or something like that?
- While I think smaller intervals than an hour would probably be fine, below ~5-10 minutes, things might get strange
- I really worked to test this in unittests thoroughly, but we should be careful in rollout. I think rolling out around midnight UTC would be preferable and monitoring results over the next hour or so.

adserver/decisionengine/backends.py
```diff
@@ -32,6 +33,10 @@ def __init__(self, request, placements, publisher, **kwargs):
         :param placements: possible positions for the ad to go
         :param kwargs: Any additional possible arguments for the backend
         """
+        # Use the global pacing
+        # This spaces the flight over the whole flight by interval
+        self.interval = settings.ADSERVER_PACING_INTERVAL
```
ericholscher, line 38: We should presumably make this configurable by flight, right? And make the setting a default? I don't think everyone will want it to be the same.

That also makes the deploy less scary, because we can just change 1 flight to the new interval.

adserver/decisionengine/backends.py
```diff
@@ -367,14 +372,15 @@ def select_flight(self):
...
                     (
-                        (flight.clicks_needed_today() > 0),
-                        (flight.views_needed_today() > 0),
+                        (flight.clicks_needed_today(self.interval) > 0),
+                        (flight.views_needed_today(self.interval) > 0),
                     )
                 ):
                     # NOTE: takes into account views for CPM ads
                     # Takes eCPM (CTR * CPC for CPC ads) into account
                     weighted_clicks_needed_today = flight.weighted_clicks_needed_today(
```
ericholscher, line 381: Yea, definitely feels like the code needs a bit of a refactor/rename to take intervals into account.

adserver/decisionengine/backends.py
```diff
@@ -273,8 +278,8 @@ def filter_flight(self, flight):
         if not flight.show_on_domain(self.url):
             return False
 
-        # Skip if there are no clicks or views needed today (ad pacing)
-        if flight.weighted_clicks_needed_today() <= 0:
+        # Skip if there are no clicks or views needed today/this interval (ad pacing)
+        if flight.weighted_clicks_needed_today(interval=self.interval) <= 0:
```
ericholscher, line 282: Seems like passing in `flight.interval` would be simple here? With the model field being defaulted to the setting?

ericholscher (approved):
This looks like a good initial implementation. I do think we're gonna want to have it be more configurable though, both for ease of deployment and as a general feature for a user/flight. I'm fine if that's a v2 addition, but I do think it would make things easier to do it up front, and roll it out slowly.

davidfischer:
Overall, I like the idea of a per-flight interval. On the surface, that feels like it should be fine. Let me try it out.

adserver/decisionengine/backends.py
davidfischer: I'm considering renaming all those functions (`*_needed_today`). Thoughts?

davidfischer:
I refactored to per-flight intervals.

adserver/models.py
```diff
@@ -686,7 +686,11 @@ class Flight(TimeStampedModel, IndestructibleModel):
     HIGHEST_PRIORITY_MULTIPLIER = 1000000
     LOWEST_PRIORITY_MULTIPLIER = 1
 
-    DEFAULT_INTERVAL = 60 * 60 * 24  # 24 hours
+    # The pacing interval in seconds used to pace flights evenly.
+    # For example, with a 1000 click flight over 10 days,
+    # there will be 10 intervals (each 1 day) and we'll aim for 100 clicks per day
+    # Shorter intervals (eg. an hour) spread flights more evenly across geographies
+    DEFAULT_PACING_INTERVAL = 60 * 60 * 24  # 24 hours
```
ericholscher, line 693: I think we probably want this to be a setting still? but maybe not?

ericholscher (approved):
This looks really nice 💯. I think eventually we'll want something even smarter here (eg. `pacing_method`), but this is a great start down that road!

adserver/models.py
davidfischer: I think it will require a migration since the default is used as a DB default.

Issue #692, opened 2023-01-23
Docker container size and ML requirements


mattishaden:
I know that the Dockerfiles are not maintained, but just checking in here for some pointers. When building the containers, it becomes very big (6,8 GB) because of the pip requirements for the keyword/topic analyzer. I saw the ML stuff were moved out to a external repository, are the corresponding packages still needed in `requirements/`?

We're using a modified Dockerfile, but testing with `/local/django/Dockerfile` produces the same issue, while `/production/django/Dockerfile` results in build error.

ericholscher:
@mattishaden Thanks for the issue. I'm curious what you're doing with the ad server? 

We aren't deploying via Docker at this point, so I'm not 100% sure the best way to run this in production. We're running directly on a machine, but it does take a good bit of memory & CPU to process the modeling. 

@davidfischer can probably chime in to answer the requirements question as well, but we do still need that for running certain parts of our operations, but not sure about in this repo.

davidfischer:
You can use the ad server without our ML model at all. It uses a built-in model by default. However, the built-in stuff still requires some requirements like textacy/spacy. If you disable the analyzer entirely (that isn't currently an option but I guess it could be), then none of that is required at all.

mattishaden:
> @mattishaden Thanks for the issue. I'm curious what you're doing with the ad server?
> 
I can drop you an e-mail about this. [eric@ericholscher.com](mailto:eric@ericholscher.com) ?

We are currently trying to upgrade from v0.41.0. It runs fine except that the database keeps growing.. Separate issue, but related to the offers table.

> You can use the ad server without our ML model at all. It uses a built-in model by default. However, the built-in stuff still requires some requirements like textacy/spacy. If you disable the analyzer entirely (that isn't currently an option but I guess it could be), then none of that is required at all.

Disable the analyzer would be preferable to us (no real use case for us there), but just knowing which requirements are optional and not would help to slim down the container size.

ericholscher:
@mattishaden Yep, that's a good email.

mattishaden:
We've upgraded to v0.49.0, that's the version before the machine learning requirements start accumulating. Alpine Dockerfile seems to abandoned in 0.54.0.

What happened with [https://github.com/readthedocs/ethicalads-model](https://github.com/readthedocs/ethicalads-model) repository? Interested to know your thoughts about the ML model going forward.

[...]

PR #703, opened 2023-02-10
Make the analyzer optional

davidfischer:
- Rework the production and dev dockerfiles to make the analyzer optional. The production dockerfile is removed and the dev docker file (which we 100% support and use) will be a sample for users to create their own dockerfile.
- Tests/coverage/linting skip the analyzer module
- Analyzer dependencies moved to their own requirement file

Refs: https://github.com/readthedocs/ethical-ad-server/issues/692

## Nice to have
- **Edit: This happened.** I'd like to move `.envs/local/django` to `.envs/local/django.sample` and then ignore `.envs/local/django`. This would allow easily having local settings that are persistent without committing them (eg. enabling and disabling the analyzer).
- **Edit: Ended up just using a comment.** I'd like to have an envvar passed to the docker image build process to build with the analyzer. That would make it easier than having to comment and uncomment things in the dockerfile.


adserver/decisionengine/backends.py
```diff
@@ -112,6 +113,9 @@ def get_analyzer_keywords(self):
         if not self.url:
             return None
 
+        if "adserver.analyzer" not in settings.INSTALLED_APPS:
```
ericholscher, line 116: Is it worth logging here? Might be a bit spammy tho.

config/settings/base.py
```diff
@@ -477,10 +476,13 @@
 
 # The backend to be used by the ad server
 # for topic and keyword analysis
+# Set to `None` to disable the analyzer entirely
 ADSERVER_ANALYZER_BACKEND = env(
     "ADSERVER_ANALYZER_BACKEND",
     default="adserver.analyzer.backends.TextacyAnalyzerBackend",
 )
+if ADSERVER_ANALYZER_BACKEND:
+    INSTALLED_APPS.append("adserver.analyzer")
```
ericholscher, line 485: This is definitely cleaner 👍

docker-compose/production/django/Dockerfile
```diff
@@ -3,39 +3,52 @@
 # This Dockerfile is not currently used or maintained
 # but may be useful in setting up your own adserver instance.
 
-FROM python:3.8-alpine
+FROM ubuntu:20.04
```
ericholscher, line 6: 😍 

Should we update the comment above this? Should we actually just remove this file if we aren't using or maintaining it?

ericholscher (approved):
Much cleaner 💯

docker-compose/production/django/Dockerfile
davidfischer: I'd be OK removing it since we don't use it at all, but perhaps providing a reference is a good thing. We do provide a bit of a reference with the development one though. Mostly, I'm just trying to help any community using the server as much as possible.

adserver/decisionengine/backends.py
davidfischer: It would be *very* spammy, but if we log at debug it should be fine.

docker-compose/production/django/Dockerfile
ericholscher: Yea, I think keeping the one we're actively using is the best approach. Agreed the dev one is a good reference. 

I'm 👍 on simplifying the code & env stuff here (eg. `.envs/local/django.sample`), as you mentioned as well, to make it a bit cleaner 👍

davidfischer: @mattishaden, we are thinking of removing the production Dockerfile and just having the dev one as a reference since we are actively using and maintaining that one. This would mean you would need to maintain your own production Dockerfile. We could ensure that the dev Dockerfile is up-to-date, working, and well documented/commented. Does that seem OK?

mattishaden: Absolutely! Just having a working reference is fine with us. No errors encountered switching from Alpine to Ubuntu and the container size is acceptable.

davidfischer: Sounds good. I'll rework this PR to remove the production docker-compose examples and just use the dev Dockerfile as a reference. It will be have ways to disable the analyzer and keep the image small.

davidfischer:
This is ready for a full review.

After merging, folks will need to copy the environment sample file before using the new development Docker image:

```bash
# Copy the sample settings. The file `.envs/local/django` will be ignored
cp .envs/local/django.sample .envs/local/django
make dockerbuild
make dockerserve
```

The Dockerfile at `docker-compose/django/Dockerfile` will serve as a reference for anybody wanting to run their own ad server. The docs now point folks in this direction as well.

To disable the analyzer, comment the analyzer requirements file in the Dockerfile and make sure to set `ADSERVER_ANALYZER_BACKEND=` in `.envs/local/django`. This will ensure that the analyzer's background tasks are not run and the analyzer will not be in `INSTALLED_APPS`.

config/settings/base.py
```diff
@@ -477,10 +476,13 @@
 
 # The backend to be used by the ad server
 # for topic and keyword analysis
+# Set to `None` to disable the analyzer entirely
 ADSERVER_ANALYZER_BACKEND = env(
     "ADSERVER_ANALYZER_BACKEND",
     default="adserver.analyzer.backends.TextacyAnalyzerBackend",
```
ericholscher, line 482: I wonder if we default this to None? Seems outside of scope for a lot of use cases.

ericholscher (approved):
This is great, and really simplifies the dev setup. I think it's a solid change 👍

config/settings/base.py
davidfischer: Ya, it could default to None. We might need to rework the test cases and how they're skipped but it would be OK.

PR #706, opened 2023-02-15
Publishers should set their name in the UA

davidfischer:


ericholscher (approved):
Looks good. We should give guidance on this to append, and not replace right? We can user's agent also if we can, or another header?

adserver/api/views.py
```diff
@@ -108,6 +108,10 @@ class AdDecisionView(GeoIpMixin, APIView):
     .. http:post:: /api/v1/decision/
 
         Authentication is required for this endpoint.
+
+        Please set your publisher name in the user agent of the request
+        and set the requesting user's user agent in ``user_ua``.
```
davidfischer, line 113:
```suggestion
        and set the requesting user's user agent in ``user_ua``.
        For example, if you would hit our API with the user agent ``python-requests/2.26.0``,
        instead send something like ``python-requests/2.26.0 + YOURPUBLISHER``.
        The actual end-user's user agent should remain unchanged.
```

PR #707, opened 2023-02-21
Use optimized indexes for staff all advertiser report

davidfischer:


adserver/views.py
```diff
@@ -1040,6 +1041,7 @@ class AdvertiserReportView(AdvertiserAccessMixin, BaseReportView):
     """A report for one advertiser."""
 
     export_view = "advertiser_report_export"
+    report = AdvertiserReport
```
ericholscher, line 1044: Wonder what it would take to use the optimized index here? We could have a seperate non-default report if you wanted to see non-paid ads to make this faster? 🤔

Though I think things like fill rate might get busted with this.

adserver/reports.py
```diff
@@ -117,14 +118,16 @@ def generate(self):
 
             # Calculate advertiser cost if the offer resulted in an ad impression
             # There's no revenue for offers with no views/clicks
-            if impression.advertisement:
+            if hasattr(impression, "advertisement"):
```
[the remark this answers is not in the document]
davidfischer: The `elif` isn't the common case for all the other report classes. This code path is taken in all the subclasses except the new one in this PR.

adserver/views.py
davidfischer: It might, but I don't think it's necessary. Doing the slower report for a single advertiser should be plenty fast.

ericholscher: Yea, except when they have a ton of data (eg. RTD or some of our largest advertisers over a 3+ month time period)

davidfischer: Probably not a bad idea to implement it then.

PR #711, opened 2023-02-24
Use the optimized publisher paid index for the all publisher report

davidfischer:


adserver/reports.py
```diff
@@ -315,6 +324,20 @@ def get_index_display(self, index):
         return super().get_index_display(index)
 
 
+class OptimizedPublisherPaidReport(PublisherReport):
+
+    """
+    A report using the optimized PublisherPaidImpression index showing daily ad performance for an advertiser.
```
ericholscher, line 330: Is this daily per advertiser, or publisher? Thinking it means this:

```suggestion
    A report using the optimized PublisherPaidImpression index showing daily ad performance for a publisher.
```

ericholscher (approved):

PR #712, opened 2023-02-24
Define placement priority order

davidfischer:
We didn't actually define the placement priority in our docs other than to mention it. Nothing said what the highest priority or lowest priority was.

Confusingly, the priorities were "reversed" with priority=1 being the highest. Firstly, I don't think anybody is using this. RTD used to use this, but doesn't anymore. The ad client doesn't support it. We should consider removing or deprecating it.

However, I wanted to at least define it as I was making changes to this code.

### DIFF
--- a/adserver/api/serializers.py
+++ b/adserver/api/serializers.py
@@ -3,6 +3,7 @@ import logging
 
 from django.core.exceptions import ValidationError
 from django.core.validators import URLValidator
+from django.utils.translation import gettext_lazy as _
 from rest_framework import serializers
 
 from ..constants import ALL_CAMPAIGN_TYPES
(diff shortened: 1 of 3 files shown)


adserver/api/serializers.py
```diff
@@ -28,7 +29,13 @@ class AdPlacementSerializer(serializers.Serializer):
     div_id = serializers.CharField()
     ad_type = serializers.CharField(required=True)
     priority = serializers.IntegerField(
-        default=1, min_value=1, max_value=10, required=False
+        default=1,
+        min_value=1,
+        max_value=10,
+        required=False,
+        help_text=_(
```
ericholscher, line 36: help_text is useful! 💯 Though I don't think it's shown anywhere, at least useful internally 👍

adserver/api/views.py
```diff
@@ -63,6 +63,7 @@ class AdDecisionView(GeoIpMixin, APIView):
             The number and order must correspond to the ``div_ids``.
         :<json string priorities: An optional ``|`` delimited string of priorities for different ad types.
             The number and order matter, applying to ``div_ids`` and ``ad_types``.
+            The lowest priority is 1 and the maximum priority is 10.
```
ericholscher, line 66: Ah, here is the external bit :)

ericholscher (approved):

PR #713, opened 2023-02-24
Prioritize ads with higher CTR

davidfischer:
~~This is built off https://github.com/readthedocs/ethical-ad-server/pull/712.~~ This merges into `main` now.

- Calculate a sampled CTR for ads daily based on the previous week
- Give better performing ads slightly more weighting. The scale ramps up based on how far different the CTRs are. For example, in a flight with 2 ads, ad X with 0.1% CTR and ad Y with 0.2% CTR, the chances for X are 6/16 ~= 37.5% and the chances for Y are 10/16 ~= 62.5%. We can play a bit with these weights if we want.


adserver/decisionengine/backends.py
```diff
@@ -393,13 +393,35 @@ def select_flight(self):
...
+            0.125: 6,
+            0.150: 7,
+            0.175: 8,
+            0.200: 9,
+            0.250: 10,
+        }
+        ad_weighting = 0
+        for threshold, weight in weights.items():
+            if ad.sampled_ctr >= threshold and weight > ad_weighting:
+                ad_weighting = weight
```
ericholscher, line 413: I wonder about a sampled log message here (`if randInt(100) == 50`, or something?) -- Could be useful to sample this information without spamming ourselves. Though maybe that's better when we run the nightly task?

adserver/decisionengine/backends.py
```diff
@@ -393,13 +393,35 @@ def select_flight(self):
 
         return None
 
+    def get_ad_ctr_weight(self, ad):
+        """Apply the ad weighting factor based on the sampled CTR."""
```
ericholscher, line 397: Seems we should define what the logic of this function is here, similar to in the PR description? It's not immediately obvious what it does:

ericholscher:
```suggestion
        """Apply the ad weighting factor based on the sampled CTR.

             Give better performing ads slightly more weighting.
             The scale ramps up based on how far different the CTRs are.
             For example, in a flight with 2 ads, ad X with 0.1% CTR and ad Y with 0.2% CTR, 
             the chances for X are 6/16 ~= 37.5% and the chances for Y are 10/16 ~= 62.5%. 
              
             We can play a bit with these weights if we want.
        """
```

adserver/decisionengine/backends.py
```diff
@@ -393,13 +393,35 @@ def select_flight(self):
...
+        ad_weighting = 0
+        for threshold, weight in weights.items():
+            if ad.sampled_ctr >= threshold and weight > ad_weighting:
+                ad_weighting = weight
+
+        return ad_weighting
+
     def select_ad_for_flight(self, flight):
         """
         Choose an ad from the selected flight.
```
ericholscher, line 419:
```suggestion
        Choose an ad from the selected flight filtered requested ``self.ad_types``. 
```

I wanted to make sure this is downstream of ad_type,
since I saw a campaign with text ads doing ~4x better CTRs.
This implementation looks great though 💯

adserver/tasks.py
```diff
@@ -668,6 +670,27 @@ def calculate_publisher_ctrs(days=7):
         publisher.save()
 
 
+@app.task()
+def calculate_ad_ctrs(days=7):
+    """Calculate sampled CTRs for all active ads for the last X days."""
+    sample_cutoff = get_ad_day() - datetime.timedelta(days=days)
+
+    for ad in Advertisement.objects.filter(live=True, flight__live=True):
```
ericholscher, line 678: I wonder if we just want to filter by a baseline of views (or clicks?) in the past X days? Seems like we could get some weird outcomes here if we don't have a baseline of at least 1000 views in the period. 

I guess the weighting doesn't throw things too heavily off, but if I have 1 view and 0 clicks (or 5 views and 1 click), my CTR is not going to be representative.

ericholscher (approved):
Excited to have this! Next step is probably doing some kind of thinking around the existing publisher/advertiser boosting we're already doing. 

https://github.com/readthedocs/ethical-ad-server/blob/c7387b9db5d1b234d1b427eb570cfd5a7510f3ce/adserver/models.py#L1110-L1117

adserver/decisionengine/backends.py
ericholscher: Almost certainly needs formatting help :)

adserver/decisionengine/backends.py
davidfischer: A nightly task is probably the best for that kind of thing.

adserver/tasks.py
davidfischer: That seems reasonable to require at least 1000 views in the period.

adserver/decisionengine/backends.py
davidfischer: I added the comment with formatting.

davidfischer:
I added a couple changes:

- There's now a 1k threshold of views before we will use a sampled CTR
- Turning this on is a per-flight option with default=False

adserver/tasks.py
```diff
@@ -687,8 +687,9 @@ def calculate_ad_ctrs(days=7):
         total_views = result["total_views"] or 0
         total_clicks = result["total_clicks"] or 0
 
-        ad.sampled_ctr = calculate_ctr(total_clicks, total_views)
-        ad.save()
+        if total_views >= min_views:
+            ad.sampled_ctr = calculate_ctr(total_clicks, total_views)
```
ericholscher, line 691: Should we set it to `None` if it's below this? That would help with the "ads not being re-evaluated" problem, and at least make sure we always have fresh-ish samples.

adserver/decisionengine/backends.py
```diff
@@ -455,10 +464,11 @@ def select_ad_for_flight(self, flight):
             # The serializer has verified that the maximum value is 10
             priority = placement.get("priority", 1)
 
-            # Give more weighting to high performing ads
-            ctr_weight = self.get_ad_ctr_weight(advertisement)
+            if flight:
```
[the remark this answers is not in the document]
ericholscher:
```suggestion
            if flight.prioritize_ads_ctr:
```

I think maybe it's meant to have this?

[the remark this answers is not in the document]
davidfischer: That's definitely what I meant =)

adserver/tasks.py
davidfischer: It's set to not None. We could set it to zero but feels a bit strange to do that.

ericholscher: But for ads that got a sampled_ctr set, but then fell below the threshold. They will just have the same, old `sampled_ctr` forever, right?

davidfischer: They will until the next time they're processed. However, nothing will deprioritize an ad more than setting it to 0. I think either leave it as-is or we should lower the threshold.

ericholscher (approved):
Looks good to try it out :shipit:

PR #717, opened 2023-03-02
Add more precision to the optimized daily tables

davidfischer:
This is a debatable change and I think it's worth considering rejecting this change. The daily rollup tables exposed a slight difference in how we calculate advertiser daily spend or publisher daily earn. Due to precision, these are rounded to the nearest $0.01 on a daily basis and that value is saved in the database for reporting.

From a spend/earn perspective, I don't think it will really matter but it could lead to some off by a cent type of "issues".

To illustrate this issue, consider an $0.10 campaign over 11 days. We will always fulfill a certain number of impressions over the life of the campaign and we will aim for $0.1/11 per day. The database will store $0.01 per day when in reality we actually fulfilled $0.091 per day. If we then add up the days, it will appear as if the campaign was $0.11 instead of $0.10.

With this change, we will round to the nearest 4 decimals before storing in the database. Reporting will still round to 2 decimals but when adding up the days, it will use the full 4 decimals.


ericholscher (approved):
> To illustrate this issue, consider an $0.10 campaign over 11 days. We will always fulfill a certain number of impressions over the life of the campaign and we will aim for $0.1/11 per day. The database will store $0.01 per day when in reality we actually fulfilled $0.091 per day. If we then add up the days, it will appear as if the campaign was $0.11 instead of $0.10.

This is a great example 👍 

I think it makes sense to store the information with more precision, but I agree there's always going to be weird rounding issues when dealing with things that cost a lot less than 1 cent on a regular basis. 

I wonder if we need to think a bit more deeply around how we're fulfilling something that eg. costs $1 CPM (eg. 10 views per cent) when it serves 5 views. I think in the end most of these rough edges get removed as they're aggregated across a month, but there's definitely some weird edge cases, and I think they probably don't matter, after thinking more about it :) 

It's like Office Space all over again 😆

PR #718, opened 2023-03-02
Reduce and simplify CTR weighting

davidfischer:
- Set the max difference to 5x (0% vs 0.15%+)
- Have fewer weights overall

### DIFF
--- a/adserver/decisionengine/backends.py
+++ b/adserver/decisionengine/backends.py
@@ -400,21 +400,15 @@ class ProbabilisticFlightBackend(AdvertisingEnabledBackend):
         Give better performing ads slightly more weighting.
         The scale ramps up based on how far different the CTRs are.
         For example, in a flight with 2 ads, ad X with 0.1% CTR and ad Y with 0.2% CTR,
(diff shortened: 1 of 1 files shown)


adserver/decisionengine/backends.py
```diff
@@ -400,21 +400,15 @@ def get_ad_ctr_weight(self, ad):
...
-            0.100: 5,
-            0.125: 6,
-            0.150: 7,
-            0.175: 8,
-            0.200: 9,
-            0.250: 10,
+            0.075: 1,
+            0.100: 2,
+            0.125: 3,
+            0.150: 4,
```
ericholscher, line 411: I think we could keep the previous weights, and just divide them in half, if you want to keep the granularity.

ericholscher (approved):

adserver/decisionengine/backends.py
davidfischer: Right now it can't handle fractional weights.

ericholscher (approved):

PR #719, opened 2023-03-03
Make disabled ads more obvious

davidfischer:
Creates a separate section for disabled ads (if any).

## Screenshots

Screenshot from 2023-03-03 10-06-45
Screenshot from 2023-03-03 10-07-01


ericholscher (approved):
💯

PR #726, opened 2023-03-15
Fix a bug with the renew math

davidfischer:


ericholscher (approved):
Heh, funny bug 👍 

I do think we could be smarter with re-ups in the future, but every default has problems.

PR #729, opened 2023-03-24
Refactor the names of `*_today` methods

davidfischer:
- These methods use intervals (except `views_today`, `clicks_today`).


ericholscher (approved):
I didn't fully read through the code to understand if this changed everywhere that was needed, but seems good assuming we just did a big search/replace refactor 👍

PR #731, opened 2023-04-06
Add a filter for the dashboard home screen

davidfischer:
Add a quick filter to the home screen to find publishers/advertisers easily.

Screenshot from 2023-04-06 15-15-36


ericholscher (approved):
Pretty simple implementation, nice 💯

PR #732, opened 2023-04-07
Use the proper permissions for adding advertisers/publishers

davidfischer:
This makes the add advertiser/publisher pages use the appropriate permissions rather than just "staff".


ericholscher (approved):
Funny those little fixes when everyone on staff isn't a superuser :)

PR #735, opened 2023-04-17
Forcing an ad/campaign ignores some targeting/filtering

davidfischer:
- Ignores campaign types
- Ignores ad live status


ericholscher:
Think the conflict is because we removed the PublisherGroup logic. 👍

Issue #194, opened 2020-07-16
Add ability to filter out specific hostnames when counting paid views/clicks


ericholscher:
I had a user mention that we should probably filter out `localhost` on our side when counting paid ads & clicks. This is tricky because some of our publishers pageviews actually come through with `localhost` on their referrer because of a configuration issue on their hosting. We probably need to have a publisher-level ability to either opt in or opt out various bans, but it's definitely something we should consider as a feature.

davidfischer:
Localhost should probably be filtered out always and it can be by [blocklisting referrers](https://ethical-ad-server.readthedocs.io/en/latest/install/configuration.html#adserver-blocklisted-referrers) however that's site wide and not publisher specific.

PR #736, opened 2023-04-18
Publisher allowed domains

davidfischer:
This is the first step toward restricting which domains a publisher can show ads on. Right now, this just logs a message at warning if this setting is turned on (off by default).

However, we have more and more publishers with 2 sites and I'd like to start being a bit more systematic there.

Ref: https://github.com/readthedocs/ethical-ad-server/issues/194.


adserver/models.py
```diff
@@ -282,6 +282,17 @@ class Publisher(TimeStampedModel, IndestructibleModel):
         blank=True,
     )
 
+    # If this is blank, all domains are allowed
+    allowed_domains = models.CharField(
+        _("Allowed domains"),
+        max_length=1024,
+        help_text=_(
+            "A space separated list of domains where the publisher's ads can appear"
```
ericholscher, line 290: Do we allow wildcards or similar here? Or is it a `if in` check or similar? Could be useful to be more specific so we know what to include, or is it an exact match?

ericholscher (approved):
This seems like a reasonable first step, but will likely need to be a bit smarter. There's gonna be a bunch of edge cases where folks have a bunch of subdomains or similar they serve ads on, but good to start being smarter here 💯

adserver/models.py
davidfischer: I think we may need those at some point. Before we implement it though, I'd like to know how common that is.

PR #738, opened 2023-05-02
Prioritize overdue flights

davidfischer:
The more overdue, the more prioritized they will be.

This is a pretty naive implementation where each day overdue multiplies the priority.

### DIFF
--- a/adserver/models.py
+++ b/adserver/models.py
@@ -1056,6 +1056,12 @@ class Flight(TimeStampedModel, IndestructibleModel):
         # (eg. 23 hours, 59 minutes, 59 seconds, 999ms)
         return max(0, int(remaining_seconds / self.pacing_interval)) + 1
 
+    def days_overdue(self):
+        """Number of days (NOT intervals) past the end date on the flight."""
+        # A negative number means the flight is not overdue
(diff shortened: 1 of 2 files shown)


ericholscher (approved):
I wonder if this is enough of a multiplier. In the past they have been quite overshadowed, so a *5 muliplied won't do much. I wonder about doing something like `self.days_overdue() * self.days_overdue()` as a multiplier, to make it get exponentially larger the later it is?

davidfischer:
> I wonder about doing something like `self.days_overdue() * self.days_overdue()` as a multiplier, to make it get exponentially larger the later it is?

I like the idea of scaling up over time but I think linear may be enough. Squaring is probably too much but maybe `^ 1.25` would be enough.

davidfischer:
Actually I ran a few simulations and I think `^1.5` is a good improvement. I'll make that change.

ericholscher:
That works for me. I think more than linear is definitely good for cases where it's getting quite outcompeted. 💯

davidfischer:
> `^1.5` works for me. I think more than linear is definitely good for cases where it's getting quite outcompeted. 100

I think any boost that increases over time such as linear will probably be enough. However, I think even `^1.5` will *definitely* be enough in all cases.

PR #741, opened 2023-05-04
Simplify daily-reports tasks in dev

davidfischer:
Run them every 5 minutes and run exactly what would run in prod.

### DIFF
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -606,6 +606,7 @@ def daily_update_publishers(day=None):
 
 @app.task(time_limit=60 * 60 * 4)
 def daily_update_reports():
+    """Update today's report data rather than the previous day."""
     day, _ = get_day()
     update_previous_day_reports(day)
 
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -605,6 +605,12 @@ def daily_update_publishers(day=None):
 
 
 @app.task(time_limit=60 * 60 * 4)
+def daily_update_reports():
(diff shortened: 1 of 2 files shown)


adserver/tasks.py
```diff
@@ -604,6 +604,12 @@ def daily_update_publishers(day=None):
             )
 
 
+@app.task(time_limit=60 * 60 * 4)
+def daily_update_reports():
+    day, _ = get_day()
+    update_previous_day_reports(day)
```
ericholscher, line 610: I don't quite understand why we have this function, instead of just calling `update_previous_day_reports` via celery? It seems like it's the same logic?

adserver/tasks.py
```diff
@@ -604,6 +604,12 @@ def daily_update_publishers(day=None):
             )
 
 
+@app.task(time_limit=60 * 60 * 4)
+def daily_update_reports():
```
ericholscher, line 608: If we need this function, we should have a docstring explaining what it's for.

ericholscher (approved):
This simplifies things, but not quite sure why we added another function instead of calling `update_previous_day_reports` directly?

adserver/tasks.py
davidfischer: The key difference is that `update_previous_day_reports` does the previous day while `daily_update_reports` does today. Should I just add a docstring?

ericholscher: But the logic is the same? `start_date, _ = get_day(day)`

PR #745, opened 2023-05-10
Stop using deprecated campaign.publishers

davidfischer:
A while back, we introduced the concept of publisher groups. These are much more useful than manually specifying the allowed publishers for an advertiser.

I audited our production data. Only a few current advertisers used this setting and I made sure they were transitioned to publisher groups.


davidfischer:
The first CI run failed here, but the tests are succeeding now. I think we have a testcase that fails after midnight UTC for some reason.

ericholscher (approved):

PR #748, opened 2023-05-25
Add rel=sponsored to ad links

davidfischer:


ericholscher (approved):
Simple enough 👍 Slightly surprised this didn't hit any test code, but guessing we don't test to this level of specificity.

PR #749, opened 2023-06-02
Show excluded publishers from the campaign

davidfischer:
There's two ways to exclude a publisher on an ad flight: for just that flight or for all flights on a campaign.


ericholscher (approved):
💯

PR #750, opened 2023-06-02
Keep only 1 year of geo/region/keyword/placement data

davidfischer:


config/settings/production.py
```diff
@@ -157,6 +157,11 @@
         # Runs on Tuesday
         "schedule": crontab(day_of_week=2, hour="6", minute="0"),
     },
+    "every-week-remove-old-report-data": {
+        "task": "adserver.tasks.remove_old_report_data",
+        # Runs on Wednesday
```
ericholscher, line 162: I wonder if it's faster to just run it nightly, so there's never more than 1 day of data to delete? I don't think it'll make a huge difference, but perhaps this comment could note why we're doing it weekly instead of nightly, if there's a specific reason?

adserver/tests/test_tasks.py
```diff
@@ -626,3 +628,47 @@ def test_daily_update_placements(self):
...
+        # Remove old aggregation data and verify they are gone
+        remove_old_report_data()
+        self.assertFalse(
+            RegionTopicImpression.objects.filter(
+                region="us-ca",
+                topic="backend-web",
+                advertisement=self.ad1,
+                date=old_date.date(),
+            ).exists()
+        )
```
ericholscher, line 674: I wonder if we should also have a newer Offer, and ensure that one doesn't get deleted?

ericholscher (approved):
Thanks for pushing this forward -- our DB will be looking nice and tidy in the next few weeks!

adserver/tests/test_tasks.py
davidfischer: There is a newer one so probably just need a separate check to ensure it isn't deleted.

config/settings/production.py
davidfischer: There's no real reason. I am not sure how fast it'll be but I assume it's going to be slow the first time and reasonably fast after that.

adserver/tests/test_tasks.py
davidfischer: I added a test for it.

ericholscher (approved):
💯

PR #751, opened 2023-06-02
Improve copying ads logic

davidfischer:
- Re-use the same form widgets as renewing flights
- Improve the logic around new names/slugs for copies. This ensures the name doesn't become "X Copy Copy Copy"
- Allow copying multiple ads at a time

### DIFF
--- a/adserver/forms.py
+++ b/adserver/forms.py
@@ -746,6 +746,57 @@ class AdvertisementForm(AdvertisementFormMixin, forms.ModelForm):
         }
 
 
+class AdvertisementCopyForm(forms.Form):
+
+    """Used by advertisers to re-use their ads."""
+
+    advertisements = AdvertisementMultipleChoiceField(
+        queryset=Advertisement.objects.none(),
+        required=False,
+        help_text=_("Copy the following advertisements"),
+    )
+
+    def __init__(self, *args, **kwargs):
+        """Add the form helper and customize the look of the form."""
+        if "flight" in kwargs:
+            self.flight = kwargs.pop("flight")
+        else:
+            raise RuntimeError("'flight' is required for the ad form")
+
+        super().__init__(*args, **kwargs)
+
+        self.fields["advertisements"].queryset = (
+            Advertisement.objects.filter(flight__campaign=self.flight.campaign)
+            .order_by("-flight__start_date", "slug")
+            .select_related()
+            .prefetch_related("ad_types")
+        )
+
+        self.helper = FormHelper()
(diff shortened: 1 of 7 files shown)


adserver/models.py
```diff
@@ -1421,15 +1421,28 @@ def __copy__(self):
...
+        # Fix up names of ads that have been copied multiple times
+        while new_name.endswith(" Copy"):
+            new_name = new_name[:-5]
+        new_slug = re.sub("-copy\d*$", "", new_slug)
+
         # Get a slug that doesn't already exist
-        new_slug = ad.slug + "-copy"
+        # This tries -copy, -copy1, -copy2, etc.
+        new_slug += "-copy"
+        digit = 0
```
ericholscher, line 1435: I wonder about using something that will change over time, like a timestamp or similar, instead of `copy`? I could see something like `datadog-us-data-science-rev1-202306` or something? Could perhaps bringing additional complexity for minimal value though..

adserver/forms.py
```diff
@@ -746,6 +746,57 @@ class Meta:
         }
 
 
+class AdvertisementCopyForm(forms.Form):
+
+    """Used by advertisers to re-use their ads."""
+
+    advertisements = AdvertisementMultipleChoiceField(
```
ericholscher, line 753: Copying multiple ads is a huge workflow improvement. This will change how I ads across flights 💯

ericholscher (approved):
Really excited about this as well!

adserver/forms.py
```diff
@@ -746,6 +746,57 @@ class Meta:
...
+        if "flight" in kwargs:
+            self.flight = kwargs.pop("flight")
+        else:
+            raise RuntimeError("'flight' is required for the ad form")
+
+        super().__init__(*args, **kwargs)
+
+        self.fields["advertisements"].queryset = (
+            Advertisement.objects.filter(flight__campaign=self.flight.campaign)
+            .order_by("-flight__start_date", "slug")
```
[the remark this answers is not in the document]
davidfischer: I don't think we want to do that in practice. For most advertisers they're just copying the most recent ads.

adserver/models.py
ericholscher: Not 100% sure, but seems more explicit than just `copy2`, to show when it was created? 🤷 It might make it more cluttered, but `Copy Copy Copy` shows users don't really care much about the name, unless they're using the slug in reports.

davidfischer: Ok, I added a change to use the date. It prunes the date when copying.

PR #752, opened 2023-06-02
Add calculator features to flight update/renew

davidfischer:
- Sold impressions or sold clicks update automatically when the budget field or the cpm/cpc fields are updated
- Budget is just a helper field and not stored on the model
- Note: Manually changing sold impressions does *NOT* update the budget or cpm/cpc

## Screenshot

Screenshot from 2023-06-02 16-39-38
(dynamic features are hard to display in a static screenshot)

### DIFF
--- a/adserver/forms.py
+++ b/adserver/forms.py
@@ -111,6 +111,13 @@ class FlightForm(FlightMixin, forms.ModelForm):
     we would need a way to let folks make adjustments and it automatically changes the price.
     """
 
+    # This is just a helper field used in JavaScript to ease flight price computation
+    # This field is *not* displayed in the Django admin because only fields in Meta.fields are displayed
+    budget = forms.IntegerField(
+        required=False,
+        label=_("Budget"),
+    )
+
     include_regions = forms.MultipleChoiceField(
         required=False,
         widget=forms.CheckboxSelectMultiple(),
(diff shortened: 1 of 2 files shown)


adserver/forms.py
```diff
@@ -111,6 +111,13 @@ class FlightForm(FlightMixin, forms.ModelForm):
     we would need a way to let folks make adjustments and it automatically changes the price.
     """
 
+    # This is just a helper field used in JavaScript to ease flight price computation
+    # This field is *not* displayed in the Django admin because only fields in Meta.fields are displayed
```
ericholscher, line 115: I do wonder about storing this instead of the number of views eventually. Though I guess there's some ambiguity around the number of views for a budget given CPM pricing, but those seem easy enough to work around 🤔

davidfischer: I've considered this a few times over the years. If I were building this over, I'd probably store a `total_budget` and then a `remaining_budget` which gets decremented every time an ad is shown or clicked. This would also let us change targeting and pricing mid-way through a flight since the price would only change going forward unlike our current logic.

ericholscher: Yea, I wasn't thinking that deeply. More that you define the flight in terms of budget, but still "account" for it in terms of views. The larger refactor here doesn't seem impossible, but likely not enough benefit to justify it at this point.

PR #753, opened 2023-06-06
Show the publisher's allowed domains

davidfischer:
The allowed domains are not yet enforcing but I think we should display this to publishers so they know they need to contact us.

We might also need to add the domains to the create publisher form.

## Screenshot

Screenshot from 2023-06-06 14-01-40

### DIFF
--- a/adserver/templates/adserver/publisher/settings.html
+++ b/adserver/templates/adserver/publisher/settings.html
@@ -23,6 +23,18 @@
     <div class="row">
 
       <div class="col-md-8 col">
+
+        {% if publisher.allowed_domains %}
+          <fieldset class="my-3 ">
+            <legend>{% trans 'Domains' %}</legend>
+            <p class="form-text">{{ publisher.allowed_domains }}</p>
+            <p class='form-text small text-muted'>
+              {% url 'support' as support_url %}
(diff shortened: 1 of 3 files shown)


adserver/templates/adserver/publisher/settings.html
```diff
@@ -23,6 +23,18 @@ <h1>{% blocktrans %}Publisher settings for {{ publisher }}{% endblocktrans %}</h
     <div class="row">
 
       <div class="col-md-8 col">
+
+        {% if publisher.allowed_domains %}
+          <fieldset class="my-3 ">
+            <legend>{% trans 'Domains' %}</legend>
+            <p class="form-text">{{ publisher.allowed_domains }}</p>
```
ericholscher, line 30: This looks a little odd, given it's a list displayed as a text field? Should we show it as a `ul`?

ericholscher (approved):

adserver/templates/adserver/publisher/settings.html
davidfischer: I switched it to a UL. This is what it looks like:
Screenshot from 2023-06-07 17-54-45


I don't feel strongly one way or the other, but happy to go with this or revert the change.

ericholscher: I think this looks more obviously like a list 💯

ericholscher (approved):

PR #755, opened 2023-06-12
Rework the payout email view/form

davidfischer:
This doesn't actually rework the email itself which I'll do in another PR.

* Show preview of the email
* Use the same front backend used elsewhere (this code predates that backend)

## Screenshot
Screenshot from 2023-06-12 16-34-13

### DIFF
--- a/adserver/staff/forms.py
+++ b/adserver/staff/forms.py
@@ -1,7 +1,6 @@
 """Views for the administrator actions."""
 import logging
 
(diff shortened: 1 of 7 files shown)


ericholscher (approved):
Looks good. I rarely ever edit this email, instead I normally just create a Front draft if I need to do anything with it, but useful to have a preview to sanity check it.

PR #760, opened 2023-06-23
Don't show publisher house advertisers on homescreen

davidfischer:
You can still get to these through the Django admin if needed but they're only very rarely needed


ericholscher (approved):
Thanks for grabbing this one. Simple enough 👍

PR #761, opened 2023-06-28
Advertisers can request a new flight

davidfischer:
- They can base it on a previous flight
- They don't specify targeting parameters (yet)
- They can put notes or a target budget in
- Submitting the form emails support

Production note: Ensure `ADSERVER_SUPPORT_TO_EMAIL` is set in prod when this goes out.

## Screenshot
Screenshot from 2023-06-28 11-49-32

### DIFF
--- a/adserver/forms.py
+++ b/adserver/forms.py
@@ -687,7 +687,7 @@ class FlightRequestForm(FlightCreateForm):
         self.instance.campaign = (
             self.old_flight.campaign
             if self.old_flight
-            else self.advertiser.campaigns.all().first()
+            else self.advertiser.campaigns.first()
         )
 
         instance = super().save(commit)
--- a/adserver/forms.py
+++ b/adserver/forms.py
@@ -1,5 +1,6 @@
 """Forms for the ad server."""
 import logging
+from datetime import timedelta
 
 import bleach
 import stripe
@@ -353,9 +354,10 @@ class FlightCreateForm(forms.ModelForm):
 
         super().__init__(*args, **kwargs)
 
-        self.fields["campaign"].queryset = Campaign.objects.filter(
-            advertiser=self.advertiser
-        )
+        if "campaign" in self.fields:
+            self.fields["campaign"].queryset = Campaign.objects.filter(
+                advertiser=self.advertiser
+            )
         self.helper = self.create_form_helper()
 
     def create_form_helper(self):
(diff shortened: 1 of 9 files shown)


adserver/forms.py
```diff
@@ -554,6 +556,170 @@ class Meta:
...
+                    template="adserver/includes/widgets/advertisement-form-option.html",
+                ),
+                css_class="my-3",
+            ),
+            Submit("submit", _("Request a new flight")),
+            HTML(
+                "<p class='form-text small text-muted'>"
+                + str(
+                    _(
+                        "Your flight will not start automatically. Your account manager will be notified to review your ads and targeting."
```
ericholscher, line 658: Useful note 👍

adserver/forms.py
```diff
@@ -554,6 +556,170 @@ class Meta:
...
+                "sold_clicks",
+            )
+            for field in fields:
+                setattr(self.instance, field, getattr(self.old_flight, field))
+
+        # We must set the campaign
+        self.instance.campaign = (
+            self.old_flight.campaign
+            if self.old_flight
+            else self.advertiser.campaigns.all().first()
```
ericholscher, line 690:
```suggestion
            else self.advertiser.campaigns.first()
```

Don't think the `all()` is needed.

adserver/forms.py
```diff
@@ -554,6 +556,170 @@ class Meta:
...
+
+    class Meta:
+        model = Flight
+
+        fields = (
+            "name",
+            "start_date",
+            "end_date",
+        )
+        widgets = {
```
ericholscher, line 713: Interesting that this is required. Does it default to a time & date, and we just want to set it to a date?

adserver/templates/adserver/email/flight_wrapup.html
```diff
@@ -22,11 +22,13 @@
...
-<p>{% blocktrans %}
-  If you have any questions or you would like to renew your flight, please reply to this email.
-  We can run another campaign with the same creative and budget,
+{% url "flight_request" as flight_request_url %}
+<p>{% blocktrans with site_domain=site.domain %}
+  If you'd like to renew your flight, you can <a href="{{ site_domain }}{{ flight_request_url }}">request a new one</a>
+  with your desired budget and targeting details.
+  We can run another campaign with the same creatives,
   or adjust things based on your feedback around what's converting best.
+  If you have any questions, please reply to this email.
```
ericholscher, line 31:
```suggestion
  If you have any questions, you can always reply to this email.
```

adserver/views.py
```diff
@@ -446,6 +451,112 @@ def get_success_url(self):
...
+            {
+                "budget": form.cleaned_data["budget"],
+                "note": form.cleaned_data["note"],
+            },
+        )
+
+        messages.success(
+            self.request,
+            _(
+                "Successfully setup a new (non-live) %(flight)s and notified your account manager"
```
ericholscher, line 491:
```suggestion
                "Successfully setup a new draft %(flight)s and notified your account manager"
```

adserver/views.py
```diff
@@ -446,6 +451,112 @@ def get_success_url(self):
...
+        mail.send_mail(
+            _("New Flight Request - %(advertiser)s")
+            % {"advertiser": self.advertiser.name},
+            message="See HTML message",
+            from_email=settings.DEFAULT_FROM_EMAIL,
+            recipient_list=[to_email],
+            html_message=render_to_string(
+                "adserver/email/flight-request.html", context
+            ),
+        )
```
ericholscher, line 557: Do we actually want to send email here, or is this creating an object in Front dynamically? I'm fine with either approach for now, but not depending on email here could be a nice benefit eventually, similar to the inbound forms.

ericholscher (approved):
This is great. Baby steps towards self-serve. Now we just need to combine the calculator and the request flight functionality, and we should be able to expose targeting info to the user when they request a flight, and we're most of the way there! 🤔

adserver/views.py
davidfischer: I'm.... not sure. This is a bit different from the support forms. Those forms submit directly to front where this one submits to the Django backend. We could just create an email using the Front email backend (currently, this uses the normal email backend) but then every single time somebody fills out this form, there will be an email from the default Front user to `support@`. Maybe that's ok?

adserver/forms.py
davidfischer: Without this, you will not get a date widget at all. The `type=date` is what does that. We aren't using a JS date widget but rather an HTML builtin one. The pattern isn't *that* important but [improves compatibility](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#handling_browser_support) with browsers that don't support `type=date`.

davidfischer:
This error is due to https://github.com/readthedocs/ethical-ad-server/pull/764. I'll re-run the tests tomorrow (same day in CI as UTC) and it will resolve.

PR #769, opened 2023-07-11
Calculate traffic fill rate daily and enforce traffic cap

davidfischer:
- Traffic caps are enforced on country, publisher, or region
- Fill rates are calculated daily as part of our aggregations on all active, paid flights


adserver/models.py
```diff
@@ -782,6 +783,33 @@ class Flight(TimeStampedModel, IndestructibleModel):
...
+        null=True,
+        default=None,
+        validators=[TrafficFillValidator()],
+    )
+
+    # If set, any publisher, country, or region whose `traffic_fill` exceeds the cap
+    # will not be eligible to show on this campaign until they're below the cap.
+    # Format is the same as `traffic_fill` but this is set manually
+    traffic_cap = JSONField(
+        _("Traffic cap"),
```
ericholscher, line 806: Do we need a link to docs or similar around the format here in the help_text? I hit this with the JSON targeting dict, where I don't know offhand what values are supported.

adserver/models.py
```diff
@@ -974,6 +1002,31 @@ def show_to_geo(self, geo_data):
...
+            if country_traffic_fill.get(geo_data.country, 0.0) > limited_countries.get(
+                geo_data.country, 100.0
+            ):
+                return False
+
+        # Check if the region traffic cap exceeds the current fill for that region
+        if self.traffic_cap and "regions" in self.traffic_cap:
+            # pylint: disable=invalid-sequence-index
+            limited_regions = self.traffic_cap["regions"]
+            traffic_fill = self.traffic_fill or {}
```
ericholscher, line 1020: Seems like it might be cleaner to just do `if self.traffic_cap and self.traffic_fill` in initial `if` block, then we can assume this value exists, and don't execute if it isn't set.

adserver/models.py
```diff
@@ -974,6 +1002,31 @@ def show_to_geo(self, geo_data):
             ):
                 return False
 
+        # Check if the country traffic cap exceeds the current fill for that country
+        if self.traffic_cap and "countries" in self.traffic_cap:
+            # pylint: disable=invalid-sequence-index
```
ericholscher, line 1007: Weird it's raising this issue when traffic_cap is a dict. Is it because of `traffic_fill.get("countries", {})` or something? 

Seems like it shouldn't apply here: 

https://pylint.readthedocs.io/en/latest/user_guide/messages/error/invalid-sequence-index.html

adserver/tasks.py
```diff
@@ -635,6 +635,9 @@ def update_previous_day_reports(day=None):
     daily_update_uplift(start_date)
     daily_update_regiontopic(start_date)
 
+    # Updates an aggregation on each paid flight
+    update_flight_traffic_fill.apply_async()
```
ericholscher, line 639: Should we also prefix this with `daily` to match all the others? Also seems like the comment should say why this one is running another celery task instead of running it inline.

adserver/validators.py
```diff
@@ -115,3 +116,57 @@ def _validate_strs(self, words):
...
+        for slug, value in region_values.items():
+            if slug not in regions:
+                raise ValidationError(self.messages["region"], params={"value": slug})
+
+            if not isinstance(value, float) or value > 1.0 or value < 0:
+                raise ValidationError(self.messages["percent"], params={"value": value})
+
+    def _validate_publishers(self, publisher_values):
+        for publisher_slug, value in publisher_values.items():
+            if not isinstance(publisher_slug, str):
```
ericholscher, line 166: Do we actually want to validate it's a valid publisher slug? Not super important, but I did accidentally set one wrong the other day, but caught it once it was displayed back to me in the flight admin.

ericholscher (approved):
This looks great, and pretty simple to implement 💯

adserver/models.py
davidfischer: Ya, it definitely triggers a lint issue and I'm not sure why. It shouldn't.

adserver/tasks.py
davidfischer: Well, this task doesn't update "daily" numbers. It updates the lifetime of the flight. I'll add a comment about why it can be async though.

adserver/validators.py
davidfischer: We don't in the targeting validator. It would cause an extra query to do that validation.

ericholscher: I think the query is fine, since it's only validated on save, but I'm fine not validating it.

PR #772, opened 2023-07-18
Flight hard stop

davidfischer:
Some advertisers running a time sensitive event want a hard stop date for their flight.

Right now, the hard stop can only be set in the Django admin but it is displayed to the user when set. The hard stop lowers the impressions/clicks so the flight is completely fulfilled. Then the normal flight wrapup email is sent at the normal time.


adserver/models.py
```diff
@@ -720,12 +720,21 @@ class Flight(TimeStampedModel, IndestructibleModel):
...
         default=datetime.date.today,
         db_index=True,
-        help_text=_("This ad will not be shown before this date"),
+        help_text=_("This flight will not be shown before this date"),
     )
     end_date = models.DateField(
         _("End Date"),
         default=default_flight_end_date,
-        help_text=_("The target end date for the ad (it may go after this date)"),
+        help_text=_("The target end date for the flight (it may go after this date)"),
```
ericholscher, line 728:
```suggestion
        help_text=_("The estimated end date for the flight"),
```

Perhaps clearer?

adserver/tasks.py
```diff
@@ -1024,6 +1025,47 @@ def update_flight_traffic_fill():
...
+    and sending a slack notification about the remaining credit.
+
+    This should be called before `notify_of_completed_flights()`
+    so that the regular wrapup emails can be sent to advertisers.
+    """
+    today = get_ad_day().date()
+
+    for flight in Flight.objects.filter(live=True, hard_stop_date__lte=today):
+        if flight.clicks_remaining() <= 0 and flight.views_remaining() <= 0:
+            continue
```
ericholscher, line 1043: Could probably use a comment. Are we skipping this because it finished properly? 
```suggestion
        # Don't execute for flights that properly finished
        if flight.clicks_remaining() <= 0 and flight.views_remaining() <= 0:
            continue
```

adserver/tasks.py
```diff
@@ -1024,6 +1025,47 @@ def update_flight_traffic_fill():
...
+
+        flight.sold_clicks = flight.total_clicks
+        flight.sold_impressions = flight.total_views
```
ericholscher, line 1046: Do we actually want to update the sold data here? Seems like we're losing data on the original buy? Seems like we'd just mark it inactive?

adserver/tasks.py
```diff
@@ -1024,6 +1025,47 @@ def update_flight_traffic_fill():
...
+        flight.save()
+
+        value_remaining = round(flight.value_remaining(), 2)
```
ericholscher, line 1049: Isn't this going to be 0 after we updated the sold_impressions?

ericholscher (approved):
Simple enough, but not 100% sure we want to be adjusting data on the flight.

davidfischer:
After discussion, I'm thinking about having a "hard stop" boolean and use the `end_date`.

davidfischer:
I switched `hard_stop` to a boolean and worked the code into the same "completed flight" code.

adserver/models.py
```diff
@@ -1186,10 +1193,12 @@ def views_needed_this_interval(self):
 
     def clicks_needed_this_interval(self):
```
ericholscher, line 1194: Do we need to update `views_needed_this_interval` as well?

ericholscher (approved):
I like this approach, seems cleaner and doesn't lose data 👍

adserver/models.py
davidfischer: Yes it does.

PR #774, opened 2023-07-31
Copy traffic_cap when renewing a flight.

ericholscher:
I wonder if this should also be optional on the campaign as well as the flight?


agjohnson (approved):
I don't have a whole lot of context here, but seems minor enough :+1:

davidfischer:
> I wonder if this should also be optional on the campaign as well as the flight?

That's worth considering although we've definitely had cases where we want flight-specific.

davidfischer (approved):

PR #778, opened 2023-08-11
Limit classifier to 20k

davidfischer:
Longer inputs can cause massive memory usage and instability.


ericholscher (approved):
This is almost certainly the actual fix for memory usage. Do we want to increase this over time, or do we think 20k is enough? It's still a bunch of text.

davidfischer:
I think 20k is probably enough. That's many pages/screens of actual text.

PR #782, opened 2023-08-23
Pricing in server

davidfischer:
**Note:** This is branched from the Django 4 upgrade: https://github.com/readthedocs/ethical-ad-server/pull/781.

Allow advertisers to select regions and topics when requesting a flight. Based on what they specify, an estimated CPM will be shown. Prices per region/topic will be stored in a JSON field on the region model. Regions and topics will have an option whether they can be selected by advertisers when requesting a new flight.

* Currently when selecting multiple regions and topics, an average is taken. So if there's 2 regions selected and 2 topics selected, the price will be an average of the 4 combinations. This seems a bit awkward.
* Discounts are currently hard coded

## Screenshot

Screenshot from 2023-08-23 16-55-45


adserver/validators.py
```diff
@@ -7,6 +7,46 @@
...
+    code = "topic-pricing-validator"
+    limit_value = None  # required for classes extending BaseValidator
+
+    def __init__(self, message=None):
+        """Initialization for the topic pricing validator."""
+        if message:
+            self.message = message
+
+    def __call__(self, value):
+        from .models import Topic  # noqa
```
ericholscher, line 25: Guessing there's a reason this is a nested import, but probably good to note it here?

assets/src/views/flight-request.js
```diff
@@ -0,0 +1,28 @@
+const ko = require('knockout');
+import { getFlightPrice, getDiscount } from "../libs/flight-utils";
+
+
+function FlightRequestViewModel(method) {
+  this.budget = ko.observable($("#id_budget").val());
+  this.regions = ko.observable([...document.querySelectorAll("form [name='regions']:checked")].map(function (node) { return node.value }));
+  this.topics = ko.observable([...document.querySelectorAll("form [name='topics']:checked")].map(function (node) { return node.value }));
+
+  this.pricing = JSON.parse($('#data-pricing').text());
```
ericholscher, line 10: Is this used? Seems like we're loading pricing from there in `getFlightPrice` also?

ericholscher (approved):
Excited for this 🎉

assets/src/views/flight-request.js
davidfischer: No. It can be removed.

adserver/validators.py
davidfischer: The validators are used to validate the models. It would be circular without this. I'll note it.

PR #784, opened 2023-09-14
Ad image help offer

davidfischer:
- Adds a note to the advertisement form about image help
- Support form now accepts a file upload (image or not)

### Support form file upload
Screenshot from 2023-09-14 11-40-40

### Offer image support in ad form
Screenshot from 2023-09-14 11-40-50


ericholscher (approved):
Looks good -- I don't expect much here, but having the file upload on support is useful regardless.

PR #786, opened 2023-09-22
Server can control multiple publisher placements

davidfischer:
This is based on #782 so as not to have conflicting migrations.

This change allows the server to control whether a publisher is allowed to have multiple placements (which the client sends since https://github.com/readthedocs/ethical-ad-client/pull/170). When merged, this will automatically reject multiple placements and we will need to turn it on per publisher.


adserver/api/serializers.py
```diff
@@ -60,6 +60,10 @@ class AdDecisionSerializer(serializers.Serializer):
     # This purposefully doesn't use a URLField so we can disregard invalid values rather than rejecting them
     url = serializers.CharField(max_length=256, required=False)
 
+    # The placement index (0-indexed)
+    # 1 or more means there's multiple placements on this page
+    index = serializers.IntegerField(required=False, min_value=0, max_value=9)
```
ericholscher, line 65: Can we not have a max here? It seems unlikely, but I know the PSF has a use case where they're showing more than 9 images in a page.

adserver/api/views.py
```diff
@@ -296,6 +296,7 @@ def decision(self, request, data):
                 keywords=keywords,
                 campaign_types=serializer.validated_data.get("campaign_types"),
                 url=url,
+                placement_index=serializer.validated_data.get("index"),
```
ericholscher, line 299: Should we call the incoming variable `placement_index` as well to keep it consistent?

ericholscher (approved):

adserver/api/serializers.py
davidfischer: Sure. I can remove it. I'd rather have a very large max (99 or even 999) than no max at all though if that makes sense though.

adserver/api/views.py
davidfischer: https://github.com/readthedocs/ethical-ad-client/pull/171

davidfischer:
I updated based on the feedback.

ericholscher:
Looks good 👍

PR #793, opened 2023-10-05
20 keywords maximum in API

davidfischer:
API users can send additional keywords, but beyond 20 are ignored.


ericholscher (approved):
Should we log this? Probably not, since it'll happen a lot, and we don't want to spam logs I guess.

PR #795, opened 2023-10-10
Add discounts to the dashboard

davidfischer:
The discounts come in the form of Stripe coupons. The description can be edited in Stripe. Currently, we have just a 10% and 15% coupon based on the total campaign size.

They appear in:

- The create/edit flight screens (for staff). The discounts are not automatically applied yet. When a flight is renewed, the discount is renewed by default.
- They appear in the memo section of invoices
- They are displayed in the flight metadata to advertisers

## Screenshot

Screenshot from 2023-10-10 16-45-51


adserver/admin.py
```diff
@@ -670,14 +670,14 @@ def action_create_draft_invoice(self, request, queryset):
             )
 
         # https://stripe.com/docs/api/invoices/create
+        description = "Thanks for your business!"
```
ericholscher, line 673: Ha, I never noticed this before :)

ericholscher (approved):
Looks good.. So just to confirm, we aren't actually changing the pricing on the invoice, just noting the discount explicitly? That seems like a great first step. 👍

davidfischer:
> Looks good.. So just to confirm, we aren't actually changing the pricing on the invoice, just noting the discount explicitly? That seems like a great first step. 👍

Correct. The pricing doesn't change at all. I did remove the weird logic we had where flights where priced per view if the number of views was not divisible by 1000 and CPM otherwise. It's always per view or click now.

adserver/admin.py
davidfischer: I usually fill out the memo line to reference a PO or if this is "invoice 1 of 3", etc.

PR #797, opened 2023-10-16
Offers now store if they are paid ad eligible

davidfischer:
Adds two new fields to the offer table:

- `paid_eligible` - whether this offer could be a paid ad
- `is_proxy` - whether the offer is to a proxy/vpn

These were combined so we have a single migration against the offers table.

Whether an offer could be paid requires the publisher to be approved for paid campaigns and that the request didn't exclude paid campaigns.

**Note:** This does not yet change the aggregations to only aggregate on paid eligible ads. There's a couple gotchas with that that I wanted to discuss before moving forward. Most of our KPIs use the `RegionImpression` or `TopicImpression` models. We can fairly trivially change those aggregations to require `paid_eligible` but that means that unpaid (pre-approval) impressions won't even show up at all. One option is to require `paid_eligible` if and only if the publisher is approved for paid campaigns.


ericholscher (approved):
This seems like a good approach. Hopefully the migration won't be too painful.. 🤞

PR #801, opened 2023-10-27
Fix the isproxy migration to be nullable

davidfischer:


ericholscher (approved):

davidfischer:
While deploying this bug, we discovered what appears to be a bug in Django. We'll dig a bit deeper but it seems that `db_table` may not be respected when creating a migration. Specifically `./manage.py sqlmigrate adserver 0089` outputted:

```BEGIN;
--
-- Add field is_proxy to click
--
ALTER TABLE "adserver_click" ADD COLUMN "is_proxy" boolean NULL;
--
-- Add field paid_eligible to click
--
ALTER TABLE "adserver_click" ADD COLUMN "paid_eligible" boolean NULL;
--
-- Add field is_proxy to offer
--
ALTER TABLE "adserver_offer" ADD COLUMN "is_proxy" boolean NULL;
--
-- Add field paid_eligible to offer
--
ALTER TABLE "adserver_offer" ADD COLUMN "paid_eligible" boolean NULL;
--
-- Add field is_proxy to view
--
ALTER TABLE "adserver_view" ADD COLUMN "is_proxy" boolean NULL;
--
-- Add field paid_eligible to view
--
ALTER TABLE "adserver_view" ADD COLUMN "paid_eligible" boolean NULL;
COMMIT;
```

When all instances of `adserver_offer` should read `adserver_offer_2023_10` based on our envvar set `db_table`. This resulted in the migration running against the wrong table. We manually ran the migration against the right table but more investigation is required.

PR #807, opened 2023-11-20
Don't disable publishers who we recently updated to paid

ericholscher:
This was setting folks back after we re-enabled paids ads.
I'm not sure this is the cleanest way to do this,
but seems reasonable.

### DIFF
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -180,7 +180,6 @@ def daily_update_placements(day=None):
         .order_by("-total_decisions")
         .iterator()
     ):
-
         impression, _ = PlacementImpression.objects.using("default").get_or_create(
             publisher_id=values["publisher"],
             advertisement_id=values["advertisement"],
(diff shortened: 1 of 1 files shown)


adserver/tasks.py
```diff
@@ -934,8 +932,9 @@ def disable_inactive_publishers(days=60, draft_only=False, dry_run=False):
     threshold = get_ad_day() - datetime.timedelta(days=days)
     site = get_current_site(request=None)
 
+    # Don't disable publishers who we recently updated to paid
     for publisher in Publisher.objects.filter(
-        allow_paid_campaigns=True, created__lt=threshold
+        allow_paid_campaigns=True, created__lt=threshold, modified__lt=threshold
```
ericholscher, line 937: Hrm, I realize we're actually modifying every publisher with the CTR average each night... so this probably won't work.

I'll leave this PR open in case we have a better idea for how to fix this issue.

davidfischer: One possibility is to change the query that updates the publisher nightly to only update paid publishers and/or to use `update()` instead of `save()` to not update the modified time.

davidfischer:
I think we could consider closing this since #824 is merged.

PR #810, opened 2023-12-19
Region and metro targeting from CF

davidfischer:
Cloudflare can send down state/territory/region and metro data. I added this to our custom transform rule. We can use these for targeting, but I'd be hesitant to use them outside of the US as I suspect data won't be extremely reliable.

Note: campaigns that target by region should target a single country (eg. US and CA/NY/TX/FL). There could be some oddities where a region from one country has the same region code as a region in another country. Weird results could ensue.


ericholscher (approved):
Easy enough

PR #813, opened 2024-01-03
Store ad rotation status on offer

davidfischer:
This will allow us to check how rotated ads are performing relative to others. It will also let us check view rates on rotated ads.


adserver/api/views.py
```diff
@@ -157,7 +157,15 @@ class AdDecisionView(GeoIpMixin, APIView):
...
-        self, ad, placement, publisher, keywords, url, forced=False, paid_eligible=False
+        self,
+        ad,
+        placement,
+        publisher,
+        keywords,
+        url,
+        forced=False,
+        paid_eligible=False,
+        rotated=False,
```
davidfischer, line 168: Thoughts on using `kwargs` for the default args?

ericholscher: Seems fine.. just to make the function sig smaller?

adserver/models.py
```diff
@@ -2425,6 +2431,10 @@ class AdBase(TimeStampedModel, IndestructibleModel):
         null=True,
     )
 
+    # Whether this ad was rotated by the client.
+    # This is needed to measure CTR of rotated ads independently
+    rotated = models.BooleanField(_("Ad was rotated"), default=None, null=True)
```
ericholscher, line 2436: Do we want an int here to know how many rotations? Not sure we care, but might be interesting to compare at some point.

ericholscher (approved):

adserver/models.py
davidfischer: Easy enough to switch

PR #815, opened 2024-01-10
Display publisher groups if not default

davidfischer:
- After merging this, make sure to set EthicalAds and RTD as default publisher groups.


ericholscher (approved):
Looks good 👍

PR #818, opened 2024-02-02
    Initial Proof of Concept for Targeting with embeddings

ericholscher:
This is very much a draft,
but shows what creating and storing embeddings in Postgres looks like.

Doesn't implement any querying, but can be done with something like::

```
# Load shell

ADSERVER_ANALYZER_BACKEND=adserver.analyzer.backends.SentenceTransformerAnalyzerBackend ./manage.py shell_plus

# Load example data into DB

import yaml
from yaml import Loader
from adserver.analyzer.tasks import analyze_url

data = yaml.load(open("/model/assets/categorized-data.yml"), Loader)

for dat in data:
     url = dat['url']
     analyze_url(url, publisher_slug='ethicaladsio', force=True)


# Run initial test query

from pgvector.django import L2Distance
from adserver.analyzer.tasks import analyze_url

url = "https://observablehq.com/"
analyze_url(url, publisher_slug='ethicaladsio', force=True)

aurl = AnalyzedUrl.objects.get(url=url)

for url in AnalyzedUrl.objects.exclude(url=aurl.url).order_by(L2Distance('embedding', aurl.embedding))[:10]:
     print(url)
```


davidfischer (commented):
I couldn't get the migrations to run correctly even after building a new docker image. Am I missing something?

```
Running migrations:
  Applying adserver_analyzer.0003_add_embeddings...Traceback (most recent call last):
...
django.db.utils.ProgrammingError: type "vector" does not exist
LINE 1: ...rver_analyzer_analyzedurl" ADD COLUMN "embedding" vector(3) ...
```

ericholscher:
@davidfischer ah, yea. You need to enable pgvector in the DB. I forgot to note that: https://github.com/pgvector/pgvector?tab=readme-ov-file#getting-started

davidfischer:
I think there's a few problems here.

* This branch hasn't taken any of the updates from `main` since ~August before we upgraded Postgres for Django 4.2. I think we need to recreate/rebase the PR as a bunch of things are going to be off and there's going to be conflicts.
* Rather than using a 3rd party docker image using an unknown version of Postgres and an unknown version of pgvector, let's just stick with the pinned version of docker we are using (15.2) and add building the extension into the Dockerfile. Hopefully it's as easy as [adding a few steps](https://www.thestupidprogrammer.com/blog/docker-with-postgres-and-pgvector-extension/) to the Dockerfile.
* Seems fairly easy to add `migrations.RunSQL('CREATE EXTENSION IF NOT EXISTS vector;'),` to the migration (can we collapse the two migrations to one?)

adserver/analyzer/models.py
```diff
@@ -55,6 +56,8 @@ class AnalyzedUrl(TimeStampedModel):
         ),
     )
 
+    embedding = VectorField(dimensions=384, default=None, null=True, blank=True)
```
davidfischer, line 59: I suspect we will need some sort of [approximate index here](https://github.com/pgvector/pgvector-python?tab=readme-ov-file#django), but that can come later.

adserver/tasks.py
```diff
@@ -934,8 +932,9 @@ def disable_inactive_publishers(days=60, draft_only=False, dry_run=False):
     threshold = get_ad_day() - datetime.timedelta(days=days)
     site = get_current_site(request=None)
 
+    # Don't disable publishers who we recently updated to paid
     for publisher in Publisher.objects.filter(
-        allow_paid_campaigns=True, created__lt=threshold
+        allow_paid_campaigns=True, created__lt=threshold, modified__lt=threshold
```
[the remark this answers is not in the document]
davidfischer: We could change the publisher CTR calculations to only run on those where paid ads are approved. That way most publishers won't be updated nightly.

requirements/analyzer.txt
```diff
@@ -16,3 +16,7 @@ https://github.com/explosion/spacy-models/releases/download/en_core_web_md-3.4.0
 #######################################################################
 # This installs PyTorch which is ~250MB
 spacy-transformers==1.1.8
+
+
+sentence-transformers
+pgvector
```
davidfischer, line 22: Because we import this directly in `models.py`, this probably has to go in the base requirements.

davidfischer: Actually this is probably just going to be a nightmare for testing. Having a field be a postgres specific field may require our testing setup to change since our tests are run with an in-mem sqlite setup.

ericholscher: FWIW, sqlite has a similar extension: https://github.com/asg017/sqlite-vss -- but might be worth just running tests in postgres? 🤷

adserver/analyzer/models.py
ericholscher: Aye, that's definitely a next step once we start querying it.

adserver/tasks.py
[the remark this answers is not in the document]
ericholscher: Ah yea.. this must have snuck in from a PR I branched off... definitely didn't do it as part of this PR.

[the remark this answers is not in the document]
ericholscher: Will roll this back, since it's a different change.

requirements/analyzer.txt
ericholscher: Looks like there isn't an easy way to use the sqlite package in Django. Another idea I had that probably makes sense:

Break the embeddings out into their own model, with a FK or OneToOne to the AnalyzedURL? That way we could keep this all self-contained.

davidfischer: I think it's fine to have the embeddings on `AnalyzedUrl`. I think we just have to change the test skip logic ([this](https://github.com/readthedocs/ethical-ad-server/blob/6f77b2c783575feb5cee121853de4f3a0c87d778/adserver/analyzer/tests.py#L26-L28)) for the analyzer. We could ensure that `adserver.analyzer` is excluded from testing entirely and when it is tested that it uses Postgres. Thoughts?

ericholscher: Yea, that likely makes sense as well, if we don't have tests for the code currently that we'd be skipping.

davidfischer (approved):
This looks great and I think disabling the analyzer is a sensible default (especially for OSS users of our project). It might be nice to find a way to run tests on the analyzer but not a blocker for merging this.

PR #820, opened 2024-02-07
Initial work for day of the week targeting

ericholscher:
Mostly just wanted initial feedback here before I finish this work up.
Does this approach look reasonable?

The other approach I considered was having `days` be either `weekdays` or `weekends`,
which might be cleaner, but thought the flexibility was more useful of having specific days.

### DIFF
--- a/adserver/decisionengine/backends.py
+++ b/adserver/decisionengine/backends.py
@@ -1,10 +1,10 @@
 """Ad decision backends."""
-import datetime
 import logging
 import random
 
 from django.conf import settings
 from django.db import models
+from django.utils import timezone
 from user_agents import parse
 
 from ..analyzer.models import AnalyzedUrl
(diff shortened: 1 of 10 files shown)


adserver/decisionengine/backends.py
```diff
@@ -292,6 +293,10 @@ def filter_flight(self, flight):
         if flight.weighted_clicks_needed_this_interval() <= 0:
             return False
 
+        # Skip if the flight is not meant to show on these days
+        if not flight.show_to_day(timezone.now().strftime("%A").lower()):
```
davidfischer, line 297: I guess locale will always be EN_US so this should *just work*.

davidfischer (approved):
> The other approach I considered was having days be either weekdays or weekends,
which might be cleaner, but thought the flexibility was more useful of having specific days.

I think this is fine and I could conceivably see advertisers skipping Fridays or Mondays or something like that.

adserver/decisionengine/backends.py
ericholscher: Ah, good point. You can also get it to return an `int`, and use that to map to the day, if we're worried about it.

PR #829, opened 2024-02-16
Improve UX for returning publishers

davidfischer:
Give publishers a better experience than returning to a metabase dashboard with no data. This walks them through steps to get back onto EthicalAds.


ericholscher (approved):
Nice UX improvement :+1:

PR #832, opened 2024-02-19
Add initial similarity API

ericholscher:
This ports over the logic from the CLI into a DRF API view

It's a basic start for now,
but gets at least a working example for local dev.

### DIFF
--- a/adserver/analyzer/views.py
+++ b/adserver/analyzer/views.py
@@ -14,6 +14,8 @@ if "adserver.analyzer" in settings.INSTALLED_APPS:
         """
         Returns a list of similar URLs and scores based on querying the AnalyzedURL embedding for an incoming URL.
 
+        Example: http://localhost:5000/api/v1/similar/?url=https://www.gitbook.com/
+
         .. http:get:: /api/v1/embedding/
 
             Return a list of similar URLs and scores based on querying the AnalyzedURL embedding for an incoming URL
@@ -35,6 +37,11 @@ if "adserver.analyzer" in settings.INSTALLED_APPS:
 
             backend_instance = SentenceTransformerAnalyzerBackend(url)
             response = backend_instance.fetch()
+            if not response:
+                return Response(
+                    {"error": "Not able to fetch content from URL"},
+                    status=status.HTTP_404_NOT_FOUND,
+                )
             processed_text = backend_instance.get_content(response)
             analyzed_embedding = backend_instance.embedding(response)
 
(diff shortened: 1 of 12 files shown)


ericholscher:
Next steps:

* One result per domain (fix a bunch of versions being returned at the same similarity)
* Only publishers w/ paid ads
* Limit to updated in the past 90 days

requirements/development.txt
```diff
@@ -35,3 +35,7 @@ django_coverage_plugin==3.1.0
 
 # Used to build the ML model
 requests-cache==0.9.5
+
+# CORS headers
```
davidfischer, line 39: These are required in production right? Probably want both of these in `base.txt` right?

config/settings/base.py
```diff
@@ -538,3 +540,10 @@
     MIDDLEWARE.append(ADSERVER_IPADDRESS_MIDDLEWARE)
 if ADSERVER_GEOIP_MIDDLEWARE:
     MIDDLEWARE.append(ADSERVER_GEOIP_MIDDLEWARE)
+
+
+CORS_ALLOWED_ORIGINS = [
```
davidfischer, line 545: This looks like a development setting. Perhaps make the prod setting read from env?

ericholscher:
I pushed a couple changes to only return a single URL for a domain, and exclude unpaid publishers, which should make the results a bit cleaner to start.

adserver/analyzer/views.py
```diff
@@ -1 +1,92 @@
...
+            urls = []
+            for url in unfiltered_urls:
+                domain = urlparse(url.url).netloc
+                if domain not in unique_domains:
+                    unique_domains.add(domain)
+                    urls.append(url)
+
+            if not len(urls) > 3:
+                return Response(
+                    {"error": "No similar URLs found"}, status=status.HTTP_404_NOT_FOUND
```
davidfischer, line 73: I do think it's a little awkward to return HTML on success and JSON on failure but not a deal-breaker.

davidfischer (approved):
I think this is fine for an initial release and we'll refine after it goes out.

adserver/analyzer/views.py
ericholscher: Yea, there's a bunch of improvements here, just wanted to not 500 :)

config/settings/base.py
```diff
@@ -538,3 +540,10 @@
...
+    "http://localhost:8000",
+    "http://127.0.0.1:8000",
+]
+CORS_ALLOW_HEADERS = ["*"]
```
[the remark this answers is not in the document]
ericholscher: Bleh.. old comment that didn't submit.

PR #837, opened 2024-03-01
Ignore publisher mobile traffic option

davidfischer:
We can disable mobile traffic from being displayed on a publisher. This is useful when a publisher has a large mobile segment of their audience that isn't a fit for our network.

If a publisher's mobile audience is ignored, there will be a message on the publisher settings page with this information.


adserver/views.py
```diff
@@ -2023,6 +2023,20 @@ def form_valid(self, form):
         )
         return result
 
+    def get(self, request, *args, **kwargs):
+        result = super().get(request, *args, **kwargs)
+
+        if self.object.ignore_mobile_traffic:
+            messages.info(
```
ericholscher, line 2030: Is this how we're displaying other on-page notifications? Seems a bit of an odd use case for notifications vs. an in-page message. This probably works fine though.

ericholscher (approved):

adserver/views.py
davidfischer: It's awkward but I didn't want it on every single publisher page.

PR #845, opened 2024-03-14
Convert similarity API back to JSON

ericholscher:


adserver/analyzer/views.py
```diff
@@ -74,19 +73,9 @@ def get(self, request):
...
-                Text:
-                </h2>
-                <textarea style="height:100%; width:80%" disabled>
-                {processed_text}
-                </textarea>
-                """
+                {
+                    "count": len(urls),
+                    "text": processed_text[:500],
+                    "results": [[url.url, url.distance] for url in urls],
```
davidfischer, line 79: I would consider returning a dict here instead of a tuple as we'll want to return more things (description, title, etc.).

davidfischer (approved):

adserver/analyzer/views.py
ericholscher: Yea, makes sense.

ericholscher:
This is blocked on changes to the frontend, but just wanted to have a PR to do dev off of for those changes 👍

adserver/analyzer/views.py
davidfischer: Key attributes to return are url, description, title, and we could just return the favicon as the image (append `favicon.ico` to the URL domain). If desired, you could skip the image and the frontend could do that.

ericholscher:
Sounds like we're just doing dev against dummy data, so no need for this PR>

PR #846, opened 2024-03-14
Remove embedding code from public repo

ericholscher:
This code is pretty specific to our workflow,
and adds a bunch of overhead to AnalyzedUrl's.
We're going to bring this small part of the code into a private repo,
similar to our ML modeling before it.


adserver/analyzer/migrations/0005_remove_embedding.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 4.2.11 on 2024-03-14 18:53
```
davidfischer, line 1: This migration should not be run until the deploy is out already. Normally I just "deprecate" fields to be cleaned up later.

davidfischer (approved):
This seems fine for now although when the embeddings code is fleshed out we might have to make it work with this.

adserver/analyzer/tasks.py
```diff
@@ -63,46 +67,59 @@ def analyze_url(url, publisher_slug, force=False):
...
 
         analyzed_keywords = backend_instance.analyze(response)  # Can be None
-        log.debug("Keywords from '%s': %s", backend.__name__, analyzed_keywords)
+        if analyzed_keywords:
+            log.debug("Keywords from '%s': %s", backend.__name__, analyzed_keywords)
+
+            for kw in analyzed_keywords:
+                keywords.add(kw)
 
         analyzed_embedding = backend_instance.embedding(response)  # Can be None
```
ericholscher, line 76: I do wonder if it makes sense to break embedding out into it's own workflow. It's kinda grown beyond analyzers, and might make more sense outside of this flow? Not 100% sure though, since I do think we probably want to minimize the number of times we download the content.

docker-compose/django/start
```diff
@@ -5,7 +5,7 @@ set -o pipefail
 set -o nounset
 
 # Reinstall dependencies without rebuilding docker image
-pip install -r /app/requirements/development.txt
```
ericholscher, line 8: This must have snuck in, reverting it back.

adserver/analyzer/models.py
```diff
@@ -68,5 +62,31 @@ def save(self, *args, **kwargs):
...
+    )
+
     class Meta:
         unique_together = ("url", "publisher")
+
+
+class AnalyzedAd(BaseAnalyzedUrl):
+    """Analyzed keywords for a given URL."""
+
+    advertiser = models.ForeignKey(
```
davidfischer, line 85: So the idea here when we want to check if a particular flight/ad landing page URL is in the DB is to just normalize the URL (strip UTMs) and then do a DB lookup with the URL and the advertiser? That seems reasonable.

adserver/analyzer/models.py
```diff
@@ -68,5 +62,31 @@ def save(self, *args, **kwargs):
...
+        Publisher,
+        help_text=_("Publisher where this URL appears"),
+        on_delete=models.CASCADE,
+    )
+
     class Meta:
         unique_together = ("url", "publisher")
+
+
+class AnalyzedAd(BaseAnalyzedUrl):
```
davidfischer, line 82: I sort of dislike this name. `AnalyzedAd` sounds like it FKs to advertisement which it doesn't. This should probably be named `AnalyzedAdvertiserUrl` or something similar.

docker-compose.yml
```diff
@@ -28,6 +28,8 @@ services:
       # Make it so we can edit the start script dynamically,
       # for example to install dependencies
       - ./docker-compose/django/start:/start
+      # Load the ethicalads_ext code from the host, so we don't have to rebuild
+      - ${PWD}/${EA_EXT_PATH:-../ethicalads-ext}:/ethicalads-ext
```
davidfischer, line 32: What happens when this directory isn't present? Does it fail silently or loudly?

ericholscher: It seems to just load a blank directory there from what I can tell.

adserver/analyzer/tasks.py
davidfischer: Ya, I've considered this as well. I think it probably could be its own workflow. It'd be nice to have it more self-contained and it could make things more simple for anybody consuming our code but not interested in the analyzers.

PR #847, opened 2024-03-14
add analyzed ad

ericholscher:
- Convert similarity API back to JSON
- Remove embedding code from public repo
- Add AnalyzedAd model to track ads


ericholscher:
Gonna merge this `migrate-embedding`, since it seems more confusing to have 2 PRs.

PR #851, opened 2024-03-21
Fix history plugin with virtual models

ericholscher:


davidfischer (approved):

davidfischer:
Since we'll never really edit these through the admin, we could just not have history on them too.

PR #857, opened 2024-03-26
Remove embedding field on AnalyzerURL

ericholscher:
This also removes the Postgres dependency in dev,
so should allow us to run the analyzer tests again.


davidfischer (approved):
This looks fine. We have to be very careful when rolling this out as we can't run this migration before the release is out.

ericholscher:
Tests are fixed in #858

PR #858, opened 2024-03-26
Squash analyzer migrations

ericholscher:
This removes pgvector imports,
allowing us to think about running analyzer tests again.

Currently this is skipping the "heavy" analyzer tests, but re-enabling the basic analyzer tests. 

Depends on https://github.com/readthedocs/ethical-ad-server/pull/857 - can't be merged until that migration is run


ericholscher:
These tests are now properly running, and showing appropriate failures.

adserver/analyzer/backends/__init__.py
```diff
@@ -1,4 +1 @@
 """Backends for analyzing URLs for keywords and topics."""
-from .eatopics import EthicalAdsTopicsBackend  # noqa
-from .naive import NaiveKeywordAnalyzerBackend  # noqa
-from .textacynlp import TextacyAnalyzerBackend  # noqa
```
ericholscher, line 4: I didn't know why we were doing imports here, but it causes a bunch of issues if we don't have this stuff installed. Seems like a nicer UX to just use their direct paths in the settings, and not force importing them if they aren't used.

Should also save us some memory & startup time on the web servers, which don't need to run this code.

config/settings/base.py
```diff
@@ -520,7 +520,7 @@
 # By default this is empty, you can find options in ``adserver.analyzer.backends``
 ADSERVER_ANALYZER_BACKEND = env.list(
     "ADSERVER_ANALYZER_BACKEND",
-    default=[],
+    default=["adserver.analyzer.backends.naive.NaiveKeywordAnalyzerBackend"],
```
ericholscher, line 523: Not 100% sure we want this, but the requirements for the naive implementation are pretty small, basically just beautifulsoup

ericholscher:
Fixed up the tests, so now they're passing 🎉

adserver/analyzer/backends/__init__.py
davidfischer: Ok. That seems fine.

requirements/testing.txt
```diff
@@ -1,2 +1,3 @@
 # Tox will setup its own environment and so it doesn't depend on anything
 tox<4.0
+beautifulsoup4==4.11.1
```
davidfischer, line 3: Why is this needed? Shouldn't tox install everything it needs?

davidfischer (commented):
I've never [squashed migrations](https://docs.djangoproject.com/en/4.2/topics/migrations/#squashing-migrations) before. Since the previous migration files are gone (from the working copy, not from git), they will all have to be run before the squash migration, correct? So again we have to be very precise in rolling this out.

requirements/testing.txt
ericholscher: It's currently in the analyzer requirements. Probably makes sense to just move it to the base requirements?

ericholscher:
This is now unblocked.

ericholscher:
@davidfischer I might plan to merge this and deploy it this week if you 👍 it

requirements/testing.txt
davidfischer: I get it. We need this for the naive analyzer (for tests) but because we aren't installing analyzer requirements this doesn't get installed. Maybe move this one to base but I do understand the desire to have requirements per-analyzer.

config/settings/base.py
davidfischer: By putting it in the default, the analyzer is going to be defaulted to on even for people who take our OSS code. I don't think we want that.. I say we keep the `default=[]` but put this as the default in `testing.py` so the naive analyzer gets tested. Thoughts?

davidfischer (approved):

adserver/analyzer/backends/__init__.py
ericholscher: These were updated in the prod env 👍

requirements/testing.txt
ericholscher: Yea.. I'll leave it here for now, but will likely end up in some kind of analyzer base requirements or similar.

PR #863, opened 2024-03-29
Allow swapping ST models

ericholscher:
This is a WIP PR that allows us to define the model in the env.


davidfischer (approved):

davidfischer (approved):
[This model](https://www.sbert.net/docs/sentence_transformer/pretrained_models.html#multi-qa-models) looks like it gives a good blend of performance and speed. Have you tested it against our stuff to get a general opinion of how it performs?

ericholscher:
> [This model](https://www.sbert.net/docs/sentence_transformer/pretrained_models.html#multi-qa-models) looks like it gives a good blend of performance and speed. Have you tested it against our stuff to get a general opinion of how it performs?

Which model? There's three there, and we're using the smallest one. We could always try with one of the larger ones.

ericholscher:
Think this is no longer relevant, and we're moving forward with another approach.

PR #865, opened 2024-04-03
Add support for niche targeting

ericholscher:
This adds support for setting a threshold of distance,
so flights outside of that distance are not shown.

This has a few limitations,
but could be a good first step for testing out a niche targeted flight,
and see what it's CTR looks like.

### DIFF
--- a/adserver/decisionengine/backends.py
+++ b/adserver/decisionengine/backends.py
@@ -17,8 +17,6 @@ from ..models import Flight
 from ..utils import get_ad_day
 from ..utils import get_client_user_agent
 
-# from ..analyzer.utils import normalize_url
-
 
 if "adserver.analyzer" in settings.INSTALLED_APPS:
     from ..analyzer.models import AnalyzedUrl
@@ -129,7 +127,6 @@ class BaseAdDecisionBackend:
             return None
 
         normalized_url = self.url
-        # normalized_url = normalize_url(self.url)
         analyzed_url = AnalyzedUrl.objects.filter(
             url=normalized_url, publisher=self.publisher
         ).first()
@@ -311,6 +308,12 @@ class AdvertisingEnabledBackend(BaseAdDecisionBackend):
         if not flight.show_to_day(timezone.now().strftime("%A").lower()):
             return False
 
+        if hasattr(self, "niche_weights"):
+            # Skip if the flight is not meant to show on these niches
+            if not flight.show_to_niche_targeting(self.niche_weights):
+                log.debug("Flight %s does not match niche targeting", flight)
+                return False
+
         return True
 
     def select_flight(self):
@@ -382,6 +385,14 @@ class ProbabilisticFlightBackend(AdvertisingEnabledBackend):
         if flights and (self.ad_slug or self.campaign_slug):
             # Ignore priorities for forcing a specific ad/campaign
             return random.choice(flights)
+
+        if self.url and "ethicalads_ext.embedding" in settings.INSTALLED_APPS:
+            from ethicalads_ext.embedding.utils import get_niche_weights
+
+            self.niche_weights = get_niche_weights(
+                self.url,
+            )
+            log.info("Niche targeting weights: %s", self.niche_weights)
 
         for possible_flights in (
             paid_flights,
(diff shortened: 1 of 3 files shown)


davidfischer (changes requested):
I think the `INSTALLED_APPS` check should be in the constructor. We can store a `niche_weights=None` in the constructor and only fill it out if there's any candidate flights with `niche_targeting`. We want to avoid doing this unless there's flights using the targeting. I'm really worried about this ballooning our time to ad.

ericholscher:
@davidfischer I updated this to be a bit smarter, and only do any embedding queries if the flights have niche targeting, and there's an embedding for the publisher URL. I think this should be safe enough to ship, since we can disable it by just removing the targeting on a test flight, and it will only do the queries when the test flight is in the candidate flights 👍

adserver/decisionengine/backends.py
```diff
@@ -393,7 +400,26 @@ def select_flight(self):
             # Choose a flight based on the impressions needed
             flight_range = []
             total_clicks_needed = 0
+
             for flight in possible_flights:
+
+                # Apply niche targeting only when any flight has it.
+                # This is to track whether we should do expensive distance queries.
```
ericholscher, line 407: I put this here so that we only run it when we hit an eligible flight that has niche targeting, so if eg. only house flights have it, it won't run if there's a paid ad.

adserver/models.py
```diff
@@ -1005,6 +1005,17 @@ def days(self):
...
+    @property
+    def niche_targeting(self) -> float:
+        """
+        Return the niche targeting parameter for this flight.
+
+        Testing out return types here as well..
+        """
+        if not self.targeting_parameters:
+            return float(0)
+        return self.targeting_parameters.get("niche_targeting", float(0))
```
davidfischer, line 1017:
```suggestion
            return 0.0
        return self.targeting_parameters.get("niche_targeting", 0.0)
```

This is minor but I believe this syntax is preferred to calling the initializer. (eg. use `= []` vs. `= list()`)

adserver/decisionengine/backends.py
```diff
@@ -393,7 +400,26 @@ def select_flight(self):
...
+                if (
+                    flight.niche_targeting
+                    and "ethicalads_ext.embedding" in settings.INSTALLED_APPS
+                ):
+                    if not self.niche_weights:
+                        from ethicalads_ext.embedding.utils import (
+                            get_niche_weights,
+                        )  # noqa
+
+                        self.niche_weights = get_niche_weights(
```
davidfischer, line 417: Unless I'm reading this wrong, won't this get the distance for every advertiser? It'd be really nice if the embedding was just cached on the flight (or directly linked with a FK so `select_related` just works). Then this check would just be a computation of getting the distance between the URL's embedding and the flight's embedding. Is that correct?

adserver/models.py
```diff
@@ -1222,6 +1233,18 @@ def show_on_domain(self, url):
...
 
+    def show_to_niche_targeting(self, weights):
+        """Filter flights based on distance of niche targeting"""
+
+        if self.niche_targeting:
+            if self.campaign.advertiser in weights:
+                goal_weight = self.niche_targeting
+                return weights[self.campaign.advertiser] < goal_weight
+            else:
+                return False
```
davidfischer, line 1244:
```suggestion
            return False
```

adserver/models.py
```diff
@@ -1222,6 +1233,18 @@ def show_on_domain(self, url):
 
         return True
 
+    def show_to_niche_targeting(self, weights):
+        """Filter flights based on distance of niche targeting"""
+
```
davidfischer, line 1238:
```suggestion
        """Filter flights based on distance of niche targeting."""
```

adserver/models.py
ericholscher: I really don't love this linting rule -- I find the explicit else much more readable, but alas.

adserver/decisionengine/backends.py
ericholscher: Yea, we have a few options here -- but in general a flight doesn't have a single embedding, since it can have multiple ads that each have their own. We can in theory "average" them in some fashion, but that isn't what we're doing currently.

In the future we can definitely change how this works a bit more, but for now I'm mostly just trying to reduce overall queries when we aren't doing targeting, so we can start testing it in prod and see how slow it is.

davidfischer: Alright. I think we'll have to calculate a flight embedding eventually but we can at least get a start with this.

davidfischer (approved):

PR #867, opened 2024-04-10
Minor bugfix related to publisher impression numbers

davidfischer:
This is a minor fix that didn't correctly count "null offers" (times a publisher requested an ad and we returned nothing at all). The only metric this affects is the fill rate for publishers.

To fix past data, we will need to go back and re-run `daily_update_publishers` (which is fast).


ericholscher (approved):
At least good to start getting good data.

PR #871, opened 2024-04-12
Staff fill rate fix

davidfischer:
There was an issue with the total calculation on fill rates in the staff fields on the publisher report.
They were calculated on all offers rather than paid only.


ericholscher (approved):
I don't fully understand this. Isn't fill rate including all offers? Is this for cases when a placement explicitly doesn't request a paid ad?

davidfischer:
> Isn't fill rate including all offers? Is this for cases when a placement explicitly doesn't request a paid ad?

Correct. If the publisher requests an ad (either paid or unpaid) and we send an unpaid ad, that should count against our fill rate. Yes we sent an ad, but not the kind they wanted. In some cases like countries where we don't fill paid ads, that's fine but we need this to properly measure paid ad fill rate in places where every ad should be paid.

PR #874, opened 2024-04-25
Update default from email

davidfischer:


ericholscher (approved):

config/settings/base.py
```diff
@@ -225,7 +225,7 @@
 # In development, emails are not sent and just logged to the console
 # --------------------------------------------------------------------------
 EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
-SERVER_EMAIL = "noreply@ethicalads.io"
+SERVER_EMAIL = env("SERVER_EMAIL", default="ads@ethicalads.io")
```
ericholscher, line 228: Should just be defaulted to example.com or something? We should likely just be overriding it.

davidfischer: Sure:

```suggestion
SERVER_EMAIL = env("SERVER_EMAIL", default="noreply@example.com")
```

PR #877, opened 2024-05-21
Add private webhook support app

davidfischer:
Adds an app with custom URLs under `/support/`. This is going to be used for some private webhooks (currently for processing some support and lead forms).


ericholscher (approved):
Glad to be getting more use out of the `ext` pattern now that we have it. Definitely a good example of something that certainly doesn't need to be in our OSS codebase :)

PR #879, opened 2024-05-24
Increase ad landing page URL length

davidfischer:
255 -> 1024


ericholscher (approved):

ericholscher:
Are there any other text fields we should be updating while we're at it?

PR #881, opened 2024-06-04
Automatic renewals

davidfischer:
- Advertisers can set their flight to automatically renew
- Changing the price and giving the discount are on ad reps
- Add renewal to the wrapup report
- Send a courtesy notification that their flight will renew a week before it does (currently a draft)
- Post to slack whenever an advertiser changes auto-renewal at all
- Post to slack when a flight renews automatically

## Screenshots
Screenshot from 2024-06-04 16-08-49
Screenshot from 2024-06-04 16-09-37


adserver/forms.py
```diff
@@ -658,6 +703,9 @@ def __init__(self, *args, **kwargs):
         self.fields["end_date"].help_text = _(
             "The target end date for this flight (it may go after this date)"
         )
+        self.fields["auto_renew"].help_text = _(
+            "Flights that automatically renew receive an additional 10% discount"
```
ericholscher, line 707: I do wonder about updating our published prices to assume auto-renewal, perhaps even with a toggle like many pricing pages have between monthly/annual pricing.

adserver/tasks.py
```diff
@@ -759,6 +762,50 @@ def notify_on_ad_image_change(advertisement_id):
...
+    end_date = get_ad_day().date() + datetime.timedelta(days=days_before)
+
+    for flight in Flight.objects.filter(
+        live=True,
+        auto_renew=True,
+        end_date=end_date,
+    ).select_related():
+        log.info("Notifying about flight %s auto-renewing", flight)
+        if settings.FRONT_ENABLED:
+            advertiser = flight.campaign.advertiser
```
ericholscher, line 779: I've been wondering if we should create properties to make these types of access a bit easier. I always forget the chain of models to get the data I want. Though it does hide queries, but we could do select_related smartly in the property when it nests more than 1 deep 🤔

adserver/tasks.py
```diff
@@ -759,6 +762,50 @@ def notify_on_ad_image_change(advertisement_id):
...
+            site = get_current_site(request=None)
+
+            to_addresses = [
+                u.email
+                for u in advertiser.user_set.all()
+                if u.notify_on_completed_flights
```
ericholscher, line 785: This is kind of a different use of this setting, but probably fine.

adserver/tasks.py
```diff
@@ -759,6 +762,50 @@ def notify_on_ad_image_change(advertisement_id):
...
+                "flight": flight,
+                "advertiser": advertiser,
+            }
+
+            with mail.get_connection(
+                settings.FRONT_BACKEND,
+                sender_name=f"{site.name} Flight Tracker",
+            ) as connection:
+                message = mail.EmailMessage(
+                    _("Advertising flight renewing - %(name)s") % {"name": site.name},
```
ericholscher, line 799:
```suggestion
                    _("Your flight will renew automatically - %(name)s") % {"name": advertiser.name},
```

Isn't EthicalAds in the from address? I'd perhaps expect this to be the company name, not our name.

adserver/tasks.py
```diff
@@ -818,6 +865,61 @@ def notify_of_completed_flights():
...
+
+            while Flight.objects.filter(slug=new_flight_slug).exists():
+                random_char = get_random_string(1)
+                new_flight_slug = slugify(f"{new_flight_slug}-{random_char}")
+
+            new_flight = Flight(
+                name=new_flight_name,
+                slug=new_flight_slug,
+                start_date=timezone.now().today(),
+                end_date=timezone.now().today() + (flight.end_date - flight.start_date),
```
ericholscher, line 886: Shouldn't this get the next month or 3 months, and subtract one day to get the end of the month? Seems like we mostly just want these to line up with month boundaries, not a specific number of days. But maybe that's complicated also? 🤷

adserver/tasks.py
```diff
@@ -818,6 +865,61 @@ def notify_of_completed_flights():
...
+                "targeting_parameters",
+                "priority_multiplier",
+                "traffic_cap",
+                "discount",
+                "cpc",
+                "cpm",
+                "sold_clicks",
+                "sold_impressions",
+                "campaign",
+                "auto_renew",  # New flight will also auto-renew
```
ericholscher, line 901: This should probably be a constant, so we're keeping this consistent across copying.

adserver/templates/adserver/email/flight-renewing.html
```diff
@@ -0,0 +1,28 @@
...
+
+  {% endblocktrans %}
+</p>
+
+<p>{% blocktrans %}Thanks again for advertising with {{ site_name }}!{% endblocktrans %}</p>
+
+{% url "account" as notification_settings_url %}
+<p>
+  <small>{% blocktrans with site_name=site.name %}You are receiving this email because you run advertising with {{ site_name }}.{% endblocktrans %} </small>
+  <small>{% blocktrans with site_domain=site.domain %}Adjust your <a href="{{ site_domain }}{{ notification_settings_url }}?ref=email-renewal">notification settings</a>{% endblocktrans %} in our dashboard.</small>
```
ericholscher, line 25: This should perhaps be in the base template as a footer of some kind?

ericholscher (approved):
This is a great next step. I do worry about managing the pricing changes with autorenewal, but we can tune that as we see if more people do this.

adserver/forms.py
davidfischer: I would make it a toggle. I think when this rolls out we also update the marketing pages.

adserver/tasks.py
davidfischer: EthicalAds is in the from address but I think we want EthicalAds in the subject as well.

Compare the subjects "Your flight will renew automatically - EthicalAds" vs. "Your flight will renew automatically - $ADVERTISER". One of these tells the advertiser *where* their ads are renewing and the other one tells them that their ads are renewing and reminds them of their own company name. I think the former is better for the advertiser.

adserver/tasks.py
davidfischer: I don't think it can be easily. Different places are copying different things (some are fully automatic while some have some input). Maybe there's a better way to do it but I think it's not a trivial refactor.

adserver/tasks.py
davidfischer: Matching the previous flight length will always work while attempting to align with month boundaries will have weird results if the flight ends mid-month or isn't a month/quarter in length.

adserver/tasks.py
davidfischer: I don't think 2 properties vs. 1 is a big difference to be honest and this is more explicit.

adserver/tasks.py
ericholscher: Yea, I assume they know it's from us because of the header. I guess the main value would be for anyone managing multiple accounts on EA, and on us to make it more obvious who we're talking to in the email thread, but don't feel strongly.

adserver/tasks.py
ericholscher: Sure, but this will consistently lead to weird results of `February 2025` campaign running 2 days into March or similar. Seems like we mostly want it to align with date boundaries, and we just shouldn't run campaign for shorter periods? I feel like this is definitely much less what a user would expect.

davidfischer: I don't think we can align with date boundaries easily or consistently. Should a campaign that starts on the 8th run for only 20 days if the next month is February? I acknowledge this is a bit strange but as written, the code functions pretty similarly to the defaults when a staff user hits the "renew" button.

adserver/templates/adserver/email/flight-renewing.html
davidfischer: Done.

adserver/tasks.py
davidfischer: After some discussion, let's try launching with this and we'll update it if needed.

PR #882, opened 2024-06-17
Update Python dependencies

davidfischer:
- Numpy added some backward incompatible changes in v2
- The really old allauth we are using had some incompatibilities with newer setuptools

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,32 +1,32 @@
 # Django itself and its requirements (pytz)
 Django==4.2.11
-pytz==2023.3
+pytz==2024.1
 
 # Asynchronous tasks
 celery[redis]==5.3.1
 
 # Static files
-whitenoise==6.0.0
+whitenoise==6.6.0
 
 # Handling and processing uploaded images
-Pillow==10.2.0
+Pillow==10.3.0
 
 # Our API
-djangorestframework==3.14.0
+djangorestframework==3.15.1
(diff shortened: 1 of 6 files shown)


requirements/base.txt
```diff
@@ -1,32 +1,32 @@
...
 
 # Form helpers
-crispy-bootstrap4==2022.1
-django-crispy-forms==2.0
+crispy-bootstrap4==2024.1
+django-crispy-forms==2.2
 
 # Authentication
-django-allauth==0.49.0
+django-allauth==0.63.3
```
ericholscher, line 26: I know we hit some various issues with upgrades here on RTD, but we're doing way more complex stuff there, so hopefully it'll just work.

ericholscher (approved):

ericholscher:
I'd love to see us using pip-compile similar to RTD on these deps, which makes updating this stuff pretty simple: https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html#pin-your-transitive-dependencies

davidfischer:
> I'd love to see us using pip-compile similar to RTD on these deps, which makes updating this stuff pretty simple

I agree. We should switch. I don't know it at all but I guess it's time to learn.

ericholscher:
I can put together a PR with it :+1:

ericholscher:
https://github.com/readthedocs/ethical-ad-server/pull/884

PR #884, opened 2024-06-20
Handle all requirements with pip-tools

ericholscher:
This adds a short little update.sh script which runs pip-compile for each file.

This pins all our transitive dependencies,
which is really nice for reproducible deployments,
so we'll never get broken from a dependency breaking in the middle.


davidfischer (approved):
Looks like a great change!

PR #893, opened 2024-07-10
Fixed a minor dependency consistency issue

davidfischer:
I basically just pinned a few dev dependencies and re-ran `./requirements/update.sh`

PR #895, opened 2024-07-10
Upgrade all dependencies

ericholscher:
This uses pip-compile -U to update all deps to their latest versions.


requirements/base.txt
```diff
@@ -66,39 +76,49 @@ django-environ==0.11.2
     # via -r base.in
 django-extensions==3.2.3
     # via -r base.in
-django-ratelimit==2.0.0
+django-ratelimit==4.1.0
```
davidfischer, line 79: We either have to pin to `<4` or switch how we're importing: https://django-ratelimit.readthedocs.io/en/latest/upgrading.html#from-3-x-to-4-0

davidfischer:
I fixed a few test issues and implemented the #893 change in here.

Replaces #893.

ericholscher:
@davidfischer Thanks for cleaning up the tests here. I think we should be good to move forward here, but this release will definitely require a good bit more testing. I know in the past we've hit some issues upgrading Sentry & New Relic as well, which will be harder to test in dev. 

Probably makes sense to merge this and test it in dev for a week or so before going fully live, and then doing strong QA on it. If we don't have any short-term changes to make, I think it should be good to merge now.

davidfischer:
> Probably makes sense to merge this and test it in dev for a week or so before going fully live, and then doing strong QA on it. If we don't have any short-term changes to make, I think it should be good to merge now.

That sounds good to me. I have run the full test suite and ran it in dev, but I agree. Having a PR with a bunch of upgrades was causing some work slowdown because the env was in a weird state so I'm glad to have this merged.

PR #897, opened 2024-07-17
Update the ad client dependency

davidfischer:
- Needed for newer styling from the ad css


ericholscher (approved):
Interesting... is this not pinned in the package.json?

davidfischer:
It's pinned to a commit in the lockfile. It won't get updated as long as it's in the lockfile.

PR #905, opened 2024-08-14
Find competitor JS with analyzer

davidfischer:


adserver/analyzer/backends/base.py
```diff
@@ -15,6 +15,15 @@
 class BaseAnalyzerBackend:
     """Base class that all analyzers should extend."""
 
+    # CSS selectors that attempt to find competing ad networks
+    # This is a predictor of publishers moving off the platform
+    COMPETITORS_SELECTORS = (
```
ericholscher, line 20: This feels like something that should probably live in our private `-ext` repo, since I don't know how relevant this is for the OSS codebase? Or at least come from a ENV variable, and not be hard coded.

ericholscher (approved):

adserver/analyzer/backends/base.py
davidfischer: I suppose I can move it to -ext and create a separate analyzer backend for this specifically

PR #906, opened 2024-08-16
First flight launched notification

davidfischer:
Creates a front draft when the first flight is launched for an advertiser. Also renames the user field for flight notifications to be more obvious.

I do wonder if we should just have this notification for *all* flights instead of just the first one


adserver/tasks.py
```diff
@@ -763,6 +763,61 @@ def notify_on_ad_image_change(advertisement_id):
...
+                "flights": flights,
+                "advertiser": advertiser,
+            }
+
+            with mail.get_connection(
+                settings.FRONT_BACKEND,
+                sender_name=f"{site.name} Flight Tracker",
+            ) as connection:
+                message = mail.EmailMessage(
+                    _("Advertising launched - %(name)s") % {"name": site.name},
```
ericholscher, line 811: This should probably say "Flight" or "Campaign". I generally use Campaign since Flight is pretty jargony and specific, and Campaign is pretty industry standard, and could have included multiple flights?

adserver/templates/adserver/email/flights-launched.html
```diff
@@ -0,0 +1,30 @@
+{% extends 'adserver/email/advertiser-base.html' %}
+{% load i18n %}
+
+
+{% block content %}
+<p>{% blocktrans with advertiser_name=advertiser.name %}{{ advertiser_name }} team,{% endblocktrans %}</p>
+
+<p>{% blocktrans with site_name=site.name %}Congrats on launching your first ad flight with {{ site_name }}!{% endblocktrans %}</p>
```
ericholscher, line 8:
```suggestion
<p>{% blocktrans with site_name=site.name %}Congrats on launching your first ad campaign with {{ site_name }}!{% endblocktrans %}</p>
```

ericholscher (approved):
Looks good, other than just some wording suggestions.

I think it could make sense to do this for all flights -- though it could get a bit spammy for folks with a bunch of different ones, but think it's generally fine. These automated emails are useful, and save us from having to do as much manual work, so 💯

adserver/tasks.py
davidfischer, line 811:
```suggestion
                    _("Advertising campaign launched - %(name)s") % {"name": site.name},
```

adserver/tests/test_tasks.py
```diff
@@ -138,6 +139,31 @@ def test_calculate_ad_ctrs(self):
...
+
+        self.flight.start_date = get_ad_day().date() - datetime.timedelta(days=1)
+        self.flight.save()
+
+        notify_of_first_flight_launched()
+
+        # Should be one email for the flight that launched 'yesterday' now
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertTrue(
+            mail.outbox[0].subject.startswith("Advertising launched")
```
davidfischer, line 164:
```suggestion
            mail.outbox[0].subject.startswith("Advertising campaign launched")
```

PR #907, opened 2024-08-20
Option for enabling/disabling keyword/topic report link

davidfischer:
This lets us expose the keyword/topic reports to advertisers on an individual basis


agjohnson (approved):

ericholscher (approved):
Seems useful. Could also just turn them on by default for everyone? I don't know if there's much value in hiding them anymore?

davidfischer:
> Seems useful. Could also just turn them on by default for everyone? I don't know if there's much value in hiding them anymore?

I think on a lot of advertisers they're just going to generate more questions than they're worth but maybe you're right. A lot of advertisers will get no value if they target one topic.

ericholscher:
Yea, I think in general folks aren't really looking too much at these reports. Honestly if they ask questions, it's probably a good to engage with them to explain it better. It shows off a lot of our differentiation around targeting, so probably good to expose it more.

davidfischer:
You've convinced me. I'll rework this to just turn it on for everyone.

PR #911, opened 2024-09-05
Filter get_niche_weights by flights

ericholscher:
This should reduce the set of things we have to filter over,
and hopefully save memory.

This won't scale cleanly to multiple flights with niche_targeting,
and will require a refactor,
but should safe a bunch of processing on the query for now on unrelated Advertisers.

### DIFF
--- a/adserver/decisionengine/backends.py
+++ b/adserver/decisionengine/backends.py
@@ -407,15 +407,12 @@ class ProbabilisticFlightBackend(AdvertisingEnabledBackend):
                     # We have to do this here,
                     # so we can filter by the weight in the filter_flight call below
                     # TODO: Handle filtering by multiple flights in the future
-                    if not self.niche_weights:
-                        from ethicalads_ext.embedding.utils import (  # noqa
-                            get_niche_weights,
-                        )
+                    from ethicalads_ext.embedding.utils import (  # noqa
+                        get_niche_weights,
+                    )
 
-                        self.niche_weights = get_niche_weights(
-                            self.url, flights=[flight]
-                        )
-                        log.info("Niche targeting weights: %s", self.niche_weights)
+                    self.niche_weights = get_niche_weights(self.url, flights=[flight])
+                    log.info("Niche targeting weights: %s", self.niche_weights)
 
                 # Handle excluding flights based on targeting
                 if not self.filter_flight(flight):
--- a/adserver/decisionengine/backends.py
+++ b/adserver/decisionengine/backends.py
@@ -402,21 +402,20 @@ class ProbabilisticFlightBackend(AdvertisingEnabledBackend):
                 flight for flight in flights if flight.niche_targeting
             ]
 
-            if flights_with_niche_targeting:
-                # Apply niche targeting only when any flight has it.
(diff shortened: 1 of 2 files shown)


adserver/decisionengine/backends.py
```diff
@@ -404,15 +404,15 @@ def select_flight(self):
...
+                    # TODO: Handle filtering by multiple flights in the future
+                    from ethicalads_ext.embedding.utils import (  # noqa
+                        get_niche_weights,
+                    )
 
-                        self.niche_weights = get_niche_weights(
-                            self.url,
-                        )
-                        log.info("Niche targeting weights: %s", self.niche_weights)
+                    self.niche_weights = get_niche_weights(self.url, flights=[flight])
```
davidfischer, line 414: This may be fine for a test but this will quickly fail to scale and might even just be wrong as written if there's 2 niche targeting flights. This method (`get_niche_weights`) currently makes 1-2 queries per call with 2 on success. I assume this is what your comment above is about. However, it seems it would be pretty easy to do something like the following at the top of the method:


```python
flights_with_niche_targeting = [flight.niche_targeting for flight in flights]
if any(flights_with_niche_targeting):
    self.niche_weights = get_niche_weights(self.url, flights=flights_with_niche_targeting)
```

Am I missing something?

ericholscher: I was mostly just trying to test out the filtering. This will work, but just do a query for each flight with it enabled. I'm fine aggregating the flights upfront, it shouldn't be hard as you mentioned.

davidfischer: I did see an issue with what I wrote above. If we have niche targeting only in house flights, it will do the aggregation even if we're selecting a paid flight since it will look at all flights. We could do this on a flight-type basis. Regardless, I think what you have is fine for a test but need to be very mindful of scaling.

ericholscher: Yea, the different types of flights and priority makes it a bit odd. We could do the basic aggregation on flight type if we wanted, as a middle ground to start.

davidfischer: > We could do the basic aggregation on flight type if we wanted, as a middle ground to start.

I think we should do that (if I'm understanding correctly). I don't want a niche targeted house flight to slow down paid ads but I'm ok if it slows down some house ads.

ericholscher: @davidfischer This will already not run on paid ads, if the only ads with niche_targeting are house ads.

davidfischer: I understand that but when we centralize the code to make it more performant, we have to ensure that stays true.

davidfischer (approved):
These changes look great. It still could result in multiple calls to `get_niche_weights` if we have a paid and house flight with niche targeting and the paid flight isn't eligible but that mostly shouldn't happen. I'm good with this for now and we can optimize later.

ericholscher:
Yea, we could always change the `possible_flights` filter to filter all flights, and even put it outside of this logic, but that would have the reserve problem (queries for house flights that might never get shown).

PR #912, opened 2024-09-09
Reduce logging, so we don't spam logs

ericholscher:


ericholscher:
Just gonna merge & hotfix this as well, to reduce spam.

PR #913, opened 2024-09-25
Fix Mailchimp email list signup

davidfischer:
We got a report that the previous URL wasn't working. It did work at one time.


davidfischer:
I think I figured it out. `https://ethicalads.us17.list-manage.com/subscribe/post?u=ca5e74de3ea2867d373058271&id=5746f18bb8` now gives an error (it used to work, I think). The URL `https://ethicalads.us17.list-manage.com/subscribe?u=ca5e74de3ea2867d373058271&id=5746f18bb8` does let people signup though. We could use that or this "landing page" which was created through Mailchimp's interface.

ericholscher (approved):
Odd that it stopped working, but this URL is nicer and seems less likely to break.

PR #915, opened 2024-09-26
Add Flights to the AnalyzedAdvertiserUrl for targeting

ericholscher:
This gives us explicit control over the URLs to target for each flight.
By default the URLs from the ads in the current flight will be added on creation,
but we can manually adjust this.


adserver/admin.py
```diff
@@ -14,6 +14,8 @@
 from djstripe.models import Invoice
 from simple_history.admin import SimpleHistoryAdmin
 
+from adserver.analyzer.models import AnalyzedAdvertiserUrl
```
davidfischer, line 17: I do wonder about this for folks who have the analyzer disabled.

davidfischer (approved):
This looks like it should be an improvement. So we can have a URL tied to the flight that is NOT one of the ad URLs, correct? That seems like the right approach as I suspect product pages will be what we want to analyze and the landing page may not always be the best fit.

adserver/admin.py
ericholscher: I think we should probably just fully migrate the analyzer over to ext, tbh. There's definitely a bunch of split brain stuff happening in this PR, and I think that's likely the cleanest thing.

PR #916, opened 2024-09-26
Automate stripe publisher payouts

davidfischer:
- Makes this just a checkbox per publisher
- Writes the transfer ID into the payout note
- Still allows handling manually if something goes wrong


ericholscher (approved):
Simple enough -- will be nice to use this next payout! I wish PayPal was this simple 🙃

PR #928, opened 2024-10-18
Build an aggregation for ad rotations

davidfischer:
## Testing

* Perform an ad rotation (even manually) but hitting the API endpoint and making sure to pass `rotations=2`.
* Aggregate the data:
    ```python
    from adserver.utils import update_previous_day_reports
    update_previous_day_reports("2024-10-18")
    ```
* Check the rotations in the admin


ericholscher (approved):
Simple enough. I really wish the aggregation logic was easier to abstract so we didn't have all of the complex query logic in each one, but so it goes.

PR #935, opened 2024-10-31
Bulk ad upload

davidfischer:
* Allow uploading ads in multiples (provide a CSV template)
* Ads can be previewed before being saved. They are validated on upload and saved into a signed field.
* After approval, the ads are saved to the database from the validated signed field

## Screenshot

Screenshot from 2024-10-31 16-32-53


adserver/forms.py
```diff
@@ -1150,6 +1157,190 @@ class Meta:
...
+        self.helper.layout = Layout(
+            Fieldset(
+                "",
+                Field("advertisements", placeholder="Upload a CSV file"),
+                css_class="my-3",
+            ),
+            Submit("submit", _("Preview ads")),
+        )
+
+    def clean_advertisements(self):
```
ericholscher, line 1207: It feels like this validation logic is likely also used in the ad form -- does it make sense to combine it into a util function?

adserver/urls.py
```diff
@@ -143,6 +144,14 @@
...
     ),
     # Advertiser management and reporting
+    path(
+        r"advertiser/advertisement-bulk-create-template.csv",
+        TemplateView.as_view(
+            template_name="adserver/advertiser/advertisement-bulk-create-template.csv",
+            content_type="text/csv",
+        ),
+        name="advertiser_bulk_create_template",
+    ),
```
ericholscher, line 154: Could also just serve this via static files? I don't feel too strongly, but this isn't a pattern I think we're using elsewhere.

adserver/forms.py
```diff
@@ -1150,6 +1157,190 @@ class Meta:
...
+                            "required_width": ad_type.image_width,
+                            "required_height": ad_type.image_height,
+                            "width": width,
+                            "height": height,
+                        },
+                    )
+
+            image_name = image_url[image_url.rfind("/") + 1 :]
+            image_path = f"images/{timezone.now():%Y}/{timezone.now():%m}/{image_name}"
+            default_storage.save(image_path, image)
```
ericholscher, line 1319: Do we want to copy the image to storage here, or after the user as saved it? I guess we want to validate we can access the file, and check sizes, so we need to download it anyway. Having a few ads saved that don't get used seems fine.

adserver/views.py
```diff
@@ -766,6 +772,161 @@ def get_success_url(self):
...
+
+    def post(self, *args, **kwargs):
+        if "advertisements" in self.request.FILES:
+            # This is step 1 of uploading the file
+            form = self.get_form()
+            if form.is_valid():
+                # CSV uploaded successfully - show preview
+                preview_ad_type = self.flight.campaign.allowed_ad_types(
+                    exclude_deprecated=True
+                ).first()
```
ericholscher, line 811: Do we want to let users set this in the future? Currently I'm assuming you can't upload text-only ads. I guess we could add it to the template in the future.

We probably want to make it explicit tho, like `image-v1`, and not just `first()`?

adserver/views.py
```diff
@@ -766,6 +772,161 @@ def get_success_url(self):
...
+        else:
+            return redirect(self.get_error_url())
+
+    def get_ad_slug(self, name):
+        slug = slugify(f"{name}")
+
+        while Advertisement.objects.filter(slug=slug).exists():
+            random_chars = get_random_string(8)
+            slug = slugify(f"{name}-{random_chars}")
+
```
ericholscher, line 890: I feel like this logic also already exists on the normal ad upload? IIRC we attached a date to it there -- should probably break that out into a util as well?

adserver/templates/adserver/advertiser/advertisement-bulk-create-template.csv
```diff
@@ -0,0 +1,2 @@
+Name,Live,Link URL,Image URL,Headline,Content,Call to Action
+Your Ad Name (required),"Whether to make your ad immediately live (required, true or false)",Your ad landing page URL (required),"Your ad image URL (required, the size must be 240*180px)",Ad headline (optional),"Ad content(required, the length of the headline, content and CTA combined must be 100 characters or less)",Ad CTA (optional)
```
ericholscher, line 2: Is there a way to specify 2 header fields on import? I'm wondering if we want to let folks keep the instructions here even after they've created the ads for import? Seems nice for "documenting" the ads on their side in the future, but not a huge requirement.

ericholscher (approved):
This looks solid! Will definitely be nice to have this.

adserver/templates/adserver/advertiser/advertisement-bulk-create-template.csv
davidfischer: I'm not sure exactly what you mean. As in, set on an ad that this was bulk-imported?

adserver/views.py
davidfischer: `first()` is not random. It is picked based on the `order` so we can control which format gets chosen. However, what if we enable a larger ad format and that's the only valid ad format for that flight? We want it to pick that one.

adserver/urls.py
davidfischer: It isn't but it keeps it self contained rather than a file that might get deleted. Or were you talking about local static files? I guess I could do that and just add another directory to `STATICFILES_DIRS`

https://github.com/readthedocs/ethical-ad-server/blob/c84adf39875cb18a829387b6b0dd72bcbf4bce87/config/settings/base.py#L272C1-L275

adserver/forms.py
davidfischer: Ya, makes sense.

adserver/urls.py
ericholscher: You could just put it in the existing static files directory, right?

davidfischer: The only static files directories we have are for images and build JS. I'll just create a new one.

adserver/forms.py
davidfischer: I moved the image/text validation onto the AdType model.

adserver/views.py
ericholscher: Yea, I know we can make `first()` return the same thing each time. But presumably we want to just define what we want it to do, instead of using this roundabout way to get an ad type?

adserver/templates/adserver/advertiser/advertisement-bulk-create-template.csv
ericholscher: Just it would be nice for this documentation to stay inside the CSV when the users are working with it. Not a huge deal, but would be nice.

adserver/views.py
davidfischer: This is just used to display the preview. It just takes the first ad type (ordered by the `order` field which we control) and uses that to display the preview. That seemed like a reasonable choice that would work here as well as on the few campaigns that use other ad types (more in the future?). Happy to do an alternative if you have a better idea.

adserver/templates/adserver/advertiser/advertisement-bulk-create-template.csv
davidfischer: I've seen this pattern on other products I've used. We could check if the 2nd line matches this file's 2nd line exactly and ignore it in that case but we want to make sure we don't ignore a row the client uses.

ericholscher: Yea, not something we need for this version necessarily, perhaps we can put this data somewhere else in the docs/UI as well?

davidfischer: The UI is probably the best place of those two. Maybe some sort of collapsable element where it can go?

PR #941, opened 2024-11-15
Add roles for advertisers and publishers

davidfischer:
- Add 3 permission levels for advertisers and publishers
- The migration will default everybody to the highest permission
- Role permissions are checked in the UI and in the backend
- Admin users can change publisher settings, and invite users.
- Managers can edit things
- Reporters can't change anything.

## Screenshot
Screenshot from 2024-11-14 16-18-07

### DIFF
--- a/adserver/templates/adserver/advertiser/advertisement-detail.html
+++ b/adserver/templates/adserver/advertiser/advertisement-detail.html
@@ -2,6 +2,10 @@
 {% load i18n %}
 {% load static %}
 {% load humanize %}
+{% load ad_extras %}
+
+
+{% advertiser_role request.user advertiser as user_advertiser_role %}
 
 
 {% block title %}{% trans 'Advertisement: ' %}{{ advertisement.name }}{% endblock %}
@@ -61,9 +65,11 @@
 </div>
 
 <div class="row mt-5">
+  {% if request.user.is_staff or user_advertiser_role  == "Admin" or user_advertiser_role  == "Manager" %}
   <div class="col">
     <a href="{% url 'advertisement_update' advertiser.slug advertisement.flight.slug advertisement.slug %}" class="btn btn-outline-primary" role="button" aria-pressed="true">{% trans 'Edit advertisement' %}</a>
   </div>
+  {% endif %}
 </div>
 
 {% endblock content_container %}
--- a/adserver/tests/test_advertiser_dashboard.py
+++ b/adserver/tests/test_advertiser_dashboard.py
@@ -25,6 +25,7 @@ from ..models import Topic
 from ..tasks import daily_update_advertisers
 from ..utils import get_ad_day
 from .common import ONE_PIXEL_PNG_BYTES
+from ..auth.models import UserAdvertiserMember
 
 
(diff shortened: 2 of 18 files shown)


adserver/tests/test_advertiser_dashboard.py
```diff
@@ -575,6 +588,19 @@ def test_flight_request_view(self):
         self.assertTrue(resp["location"].startswith("/accounts/login/"))
 
         # Regular user - access to this advertiser
+        self.client.force_login(self.user)
+
+        # Make it a reporter who can't access
+        member = UserAdvertiserMember.objects.get(user=self.user, advertiser=self.advertiser)
+        member.role = "Reporter"
```
ericholscher, line 595: I think we probably want constants for these? Not as important in tests, of course, but feels weird to have so many strings scattered around.

adserver/views.py
```diff
@@ -2421,11 +2443,15 @@ def get_queryset(self):
         self.publisher = get_object_or_404(
             Publisher, slug=self.kwargs.get("publisher_slug", "")
         )
-        return self.publisher.user_set.all()
+        return (
+            UserPublisherMember.objects.filter(publisher=self.publisher)
+            .select_related()
+            .all()
```
ericholscher, line 2449: Are we doing this so we don't have to query the relationship again? Feels a bit fiddly than just passing the users in and querying the role from something on the user? I don't feel too strongly here, but just feels a bit complex.

ericholscher (approved):
Overall this looks great. Just had a couple questions on the template tags and not using too many constants, but overall looks solid.

adserver/templates/adserver/advertiser/advertisement-detail.html
```diff
@@ -61,9 +65,11 @@ <h1>{% block heading %}{% trans 'Advertisement: ' %}{{ advertisement.name }}{% e
 </div>
 
 <div class="row mt-5">
+  {% if request.user.is_staff or user_advertiser_role  == "Admin" or user_advertiser_role  == "Manager" %}
```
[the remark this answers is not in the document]
davidfischer: One issue is that it will require a parameter (the advertiser or publisher). It's probably a good idea to do that in a method and that's a good place to cache it too. However, we'll probably still need a template tag of sorts since it accepts a parameter.

adserver/tests/test_advertiser_dashboard.py
davidfischer: Sure. There are actually constants and I should use them in the tests.

adserver/views.py
davidfischer: This is the authorized users view (a list view). It needs a list or queryset in order to for the class-based-view to do its thing. This is that queryset. It can't just be a list of users on the publisher anymore as it needs the role.

adserver/templates/adserver/advertiser/advertisement-detail.html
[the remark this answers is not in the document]
davidfischer: I added methods and attempted to simplify but we still need templatetags.

PR #954, opened 2024-12-02
Collect the domain on offer creation

davidfischer:
This just adds a migration to collect the domain (we already store the full URL). The next step will be to build an aggregation around this and a report.


ericholscher (approved):

adserver/migrations/0100_add_offer_domain.py
```diff
@@ -0,0 +1,28 @@
...
+        ('adserver', '0099_link_advertiser_guide'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='click',
+            name='domain',
+            field=models.CharField(blank=True, max_length=10000, null=True, verbose_name='Domain'),
+        ),
+        migrations.AddField(
```
ericholscher, line 18: We might need to migrate this manually because of the issue with `db_table`.

davidfischer: We might. However, the migration file is correct. Worst case we can use `sqlmigrate` and alter it.

```
./manage.py sqlmigrate adserver 0100
```

PR #955, opened 2024-12-04
Domain aggregation

davidfischer:


adserver/models.py
```diff
@@ -2498,6 +2498,30 @@ def __str__(self):
         return "Uplift of %s on %s" % (self.advertisement, self.date)
 
 
+class DomainImpression(BaseImpression):
+    """
+    Create an index of domains for each advertisement
+
+    Indexed one per ad/domain per day.
```
ericholscher, line 2505: I wonder if this should be per flight? That's how the niche targeting is scoped. Per ad will definitely increase the size, but shouldn't be too bad.

ericholscher (approved):

adserver/models.py
davidfischer: We don't have any aggregations that are currently per flight. All these aggregations are subtly different and that does make factoring common code a challenge.

ericholscher: Per ad is probably fine to start with, then.

PR #968, opened 2024-12-19
Advertiser domain report

davidfischer:
Adds an advertiser domain report where they can see the top domains for their ads. 

- Adds a note/alert if the timeframe is earlier than when data began being collected.
- We don't currently mention that some ads do not report their domain (eg. native ones) so this won't add up to the total in all cases. How should we mention that?

## Screenshot
image

### DIFF
--- a/adserver/templates/adserver/base.html
+++ b/adserver/templates/adserver/base.html
@@ -127,6 +127,16 @@
               </li>
 
               <li class="nav-item">
+                <a
+                  class="nav-link"
+                  href="{% url 'advertiser_domain_report' advertiser.slug %}"
+                >
+                  <span class="fa fa-laptop fa-fw ml-4 text-muted" aria-hidden="true"></span>
+                  <span>{% trans 'Domains' %}</span>
+                </a>
+              </li>
+
+              <li class="nav-item">
                 <a class="nav-link" href="{% url 'advertiser_users' advertiser.slug %}">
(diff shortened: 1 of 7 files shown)


adserver/templates/adserver/base.html
```diff
@@ -126,6 +126,16 @@ <h6 class="text-muted">{{ advertiser }}</h6>
                 </a>
               </li>
 
+              <li class="nav-item">
```
ericholscher, line 129: Do we want to test this behind a staff flag before rolling it out more widely?

adserver/views.py
```diff
@@ -1639,6 +1641,61 @@ def get_context_data(self, **kwargs):
...
+
+        advertiser_slug = kwargs.get("advertiser_slug", "")
+        advertiser = get_object_or_404(Advertiser, slug=advertiser_slug)
+
+        flight_slug = self.request.GET.get("flight", "")
+        flight = Flight.objects.filter(
+            campaign__advertiser=advertiser, slug=flight_slug
+        ).first()
+
+        if context["start_date"] < self.DATA_COLLECTION_START_DATE:
```
ericholscher, line 1664: This is a nice affordance.

ericholscher (approved):

adserver/templates/adserver/base.html
davidfischer: I think it's fine to just roll out. We can test it on staging but I think it should be fine.

PR #971, opened 2025-01-02
Switch to UUID7s

davidfischer:
UUID7 is a timestamp ordered UUID with some random bits. See https://www.rfc-editor.org/rfc/rfc9562.html#name-uuid-version-7. This will ensure records are timestamp ordered by the PK.


ericholscher (approved):
Simple enough. :+1: The migrations are mostly just metadata right?

davidfischer:
Correct. The migrations do not change the database. They only change how data will be created in the database post-migration.

PR #998, opened 2025-03-14
Add 2FA using allauth

davidfischer:
Most of the work here is templating and ensuring the settings are correct. There's no 2FA mandate and this doesn't enable webauthn/passkeys for now, strictly TOTP and recovery codes.

### DIFF
--- a/adserver/templates/adserver/base.html
+++ b/adserver/templates/adserver/base.html
@@ -42,6 +42,10 @@
               <span class="fa fa-key fa-fw mr-2 text-muted" aria-hidden="true"></span>
               <span>{% trans 'Change password' %}</span>
             </a>
+            <a class="dropdown-item" href="{% url 'mfa_index' %}">
+              <span class="fa fa-unlock-alt fa-fw mr-2 text-muted" aria-hidden="true"></span>
+              <span>{% trans 'MFA' %}</span>
+            </a>
             <div class="dropdown-divider"></div>
             <a class="dropdown-item" href="{% url 'support' %}">
               <span class="fa fa-envelope fa-fw mr-2 text-muted" aria-hidden="true"></span>
--- a/adserver/templates/adserver/base.html
+++ b/adserver/templates/adserver/base.html
@@ -34,17 +34,18 @@
               <span class="fa fa-user fa-fw mr-2 text-muted" aria-hidden="true"></span>
               <span>{% trans 'Account Settings' %}</span>
             </a>
-            <a class="dropdown-item" href="{% url 'api_token_list' %}">
-              <span class="fa fa-cog fa-fw mr-2 text-muted" aria-hidden="true"></span>
-              <span>{% trans 'API token' %}</span>
-            </a>
+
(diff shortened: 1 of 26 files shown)


stsewd (approved):

config/settings/base.py
```diff
@@ -70,7 +70,7 @@
     "django.contrib.humanize",
     "allauth",
     "allauth.account",
-    "allauth.socialaccount",
+    "allauth.mfa",
```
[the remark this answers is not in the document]
davidfischer: We don't plan to allow signin with github or any other social provider currently. The main users of the ad server are marketers who may not have github accounts.

adserver/templates/adserver/base.html
```diff
@@ -42,6 +42,10 @@
               <span class="fa fa-key fa-fw mr-2 text-muted" aria-hidden="true"></span>
               <span>{% trans 'Change password' %}</span>
             </a>
+            <a class="dropdown-item" href="{% url 'mfa_index' %}">
+              <span class="fa fa-unlock-alt fa-fw mr-2 text-muted" aria-hidden="true"></span>
+              <span>{% trans 'MFA' %}</span>
```
ericholscher, line 47: This should probably be spelled out? 

```suggestion
              <span>{% trans 'Multi-factor authentication (MFA)' %}</span>
```

templates/mfa/index.html
```diff
@@ -0,0 +1,69 @@
...
+  <section class="my-5" id="mfa-totp">
+    <h3>{% trans "Authenticator App" %}</h3>
+
+    {% url 'mfa_deactivate_totp' as deactivate_url %}
+    {% url 'mfa_activate_totp' as activate_url %}
+
+    {% if authenticators.totp %}
+      <p>
+        {% blocktrans trimmed %}
+          Two-factor authentication using an authenticator app is enabled.
```
ericholscher, line 28: Are we calling it 2FA or MFA? We should try to use the same language.

templates/mfa/index.html
```diff
@@ -0,0 +1,69 @@
...
+        {% endblocktrans %}
+      </p>
+      <p class="mb-0"><a href="{{ deactivate_url }}" class="btn btn-sm btn-danger">{% trans 'Deactivate MFA' %}</a></p>
```
ericholscher, line 31: There's a bunch of places where we're using this acronym, but seems ambiguous for users, and never defined in the UI?

ericholscher (approved):

templates/mfa/index.html
davidfischer: Sounds good. I'll make sure this fits. If it doesn't, I'll just be consistent in what we're calling it.

adserver/templates/adserver/base.html
davidfischer: This doesn't fit here but I spelled out "Two-factor Auth" and then defined the acronym on the 2FA detail screen.

davidfischer:
Updated for consistent verbiage and added a section in the account form with a link to 2FA

image

ericholscher (approved):

PR #1012, opened 2025-04-22
Add an optional advertiser daily cap

davidfischer:
When the daily cap is enabled, there will be a small performance degradation. However, there should be virtually no slowdown when there's no cap enabled.


adserver/models.py
```diff
@@ -1417,6 +1426,23 @@ def percent_complete(self):
...
+
+        When the daily cap is enabled (>0), then an extra query may be performed
+        for this check causing a slight performance degradation.
+        """
+        if not self.daily_cap or self.daily_cap <= 0.0:
+            return False
+
+        if self.cpm > 0 and (self.views_today() * self.cpm / 1000.0) > self.daily_cap:
+            return True
+        if self.cpc > 0 and (self.clicks_today() * self.cpc) > self.daily_cap:
```
ericholscher, line 1441: We should probably use a computed CPM/CPC helper instead of computing it here? I always get nervous spreading this logic with magic `1000` variables across the codebase.

ericholscher (approved):

adserver/models.py
davidfischer: Added

PR #1016, opened 2025-05-05
Aggregation Optimizations

davidfischer:
We have a few aggregations which run nightly to make querying the previous day's data faster. Some of these are quite expensive on the database and we're looking to move this processing out of the database. We already write the day's "Offers" to a parquet file and these aggregations can be run against that file more efficiently.

This PR starts by running only the Domain aggregation against that file and only if the day's offers are present.

While putting this together, I also noticed we're seeing very small discrepencies (~100 total impressions per day out of ~600k+) and this is due to ad views for an offer from one day occurring on the next day after the Offers have been dumped. This lowers the threshold of how long an offer can be viewed for from 4 hours -> 2 hours and makes the offer dump happen 2 hours after UTC midnight. These discrepancies should disappear.

### DIFF
--- a/adserver/models.py
+++ b/adserver/models.py
@@ -2762,8 +2762,8 @@ class Offer(AdBase):
 
     def is_old(self):
         """Checks if this offer is "old" meaning not for a currently running ad."""
-        four_hours_ago = timezone.now() - datetime.timedelta(hours=4)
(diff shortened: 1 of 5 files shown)


ericholscher:
The bits hooking it up :D

PR #1019, opened 2025-05-28
Create draft invoice for auto-renewals

davidfischer:
This change lets customers specify whether they want to receive a monthly invoice or just bill the card on file when auto-renewed. It also creates a draft invoice when the flight is renewed. I'm not in love with creating a draft invoice but at least it is explicit and creates a direct action item.

There's a few situations where we might want to modify the invoice before sending:

- Multiple flights are renewed at the same time and these should have a combined invoice.
- The customer has some special information required on the invoice beyond the info attached to the customer. VAT info can be attached but some customers require PO#s referenced.


adserver/models.py
```diff
@@ -1469,6 +1480,105 @@ def active_invoices(self):
...
+    def create_invoice(cls, flights):
+        """Create a new draft invoice for the given flights."""
+        if not settings.STRIPE_ENABLED:
+            raise ValueError(
+                _("Stripe is not configured. Please set STRIPE_SECRET_KEY.")
+            )
+
+        if not flights:
+            # No flights to create an invoice for. Don't error, just do nothing.
+            return None
```
ericholscher, line 1493: This seems a bit odd? Presumably this would be a bug?

adserver/templates/adserver/email/flight-renewing.html
```diff
@@ -17,6 +17,14 @@
   {% endblocktrans %}
 </p>
 
+{% if flight.auto_renew_payment_method %}
+<p>
+  {% blocktrans with payment_method=flight.auto_renew_payment_method|lower %}
+  For your new flight we will {{ payment_method }}.
```
ericholscher, line 23: This is likely going to be have caps -- I'd probably just make this more formal?

```suggestion
  Payment method: {{ payment_method }}.
```

ericholscher (approved):
This looks good, but I'd argue we should probably just have an option for creating real draft invoices that we can set, and generally just automatically create real ones. I think the edge cases are real, but I think the general case would be nice to just autogenerate. Alternatively the code could check for multiple renewable flights and just do a draft in that case? For PO #'s, I'm fine just editing them when needed.

PR #1021, opened 2025-05-29
Keyword impression aggregation optimization

davidfischer:
Unlike the other aggregations, this one is a bit more complicated and it might be more complex than a single query. However, there were a number of optimizations I was able to run which get the time down to ~2-3m per day for our data.

- Batch the inserts when the data is all aggregated
- Don't iterate over parts of the set that don't matter (offers w/o keywords or flights not targeted by keyword/topic)


ericholscher (approved):
Curious if this uses a lot of memory, but it's mostly ints except for the date, so should hopefully not be terrible 🙏

PR #1024, opened 2025-06-05
June 2025 Python Dependency update

davidfischer:
In addition to general updates, there was a number of additional analyzer updates and more precise pinning.

### DIFF
--- a/requirements/analyzer.in
+++ b/requirements/analyzer.in
@@ -1,20 +1,24 @@
 # Used by the keyword/topic analyzer
 beautifulsoup4
 textacy
-spacy
+spacy[transformers] >=3.8.7,< 4.0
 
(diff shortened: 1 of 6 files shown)


ericholscher:
I'd be happy to turn off dependabot as well -- those PRs just seem like noise.

davidfischer:
> This seems fine, but it seems it's adding dependencies while also updating them? Also, we should really note why we're doing an upper pin on things, otherwise we'll have no idea why/when we can update them.

I'll add notes. There isn't a lot of reason for an upper pin other than to avoid breaking changes. If a major version has breaking changes, shouldn't we be careful in taking them? The minimum pin on some of these is required because versions conflict with other modules.

PR #1035, opened 2025-06-16
Publisher payouts performance fix

davidfischer:
This improves the performance of getting what a publisher is owed for this month to date and for other periods (eg. publisher getting a payout after a few months). This will also improve the performance of the staff payout list view.

It uses the optimized aggregation we have for publisher paid impressions. As a result, it can be up to 5 minutes out of date but that's probably fine for almost all purposes.

A publisher who is getting their first payout after a long time reported this due to a dashboard 502. For that case, getting what they're owed in their first payout next month but the time period for calculating their payout is almost 2 years.


ericholscher (approved):
Fancy :)

PR #1036, opened 2025-06-17
Bugfix last analyzed date on analyzer

davidfischer:
We have some URLs in our DB that haven't been analyzed. These can be created by `daily_visited_urls_aggregation`. This causes the `analyze_url` task to fail when it is called on one of those which it currently never is automatically.

I'm going to have a separate PR to start analyzing those.

### DIFF
--- a/adserver/analyzer/tasks.py
+++ b/adserver/analyzer/tasks.py
@@ -33,6 +33,7 @@ def analyze_url(url, publisher_slug, force=False):
     - Creates or updates an entry in AnalyzedUrls table
     - Discovers keywords and topics for a URL
     """
+    skip_renanalyze_days_threshold = 30
     normalized_url = normalize_url(url)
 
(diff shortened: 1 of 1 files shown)


adserver/analyzer/tasks.py
```diff
@@ -33,6 +33,7 @@ def analyze_url(url, publisher_slug, force=False):
     - Creates or updates an entry in AnalyzedUrls table
     - Discovers keywords and topics for a URL
     """
+    skip_renanalyze_days_threshold = 30
```
ericholscher, line 36: I know we're using these magic numbers in a few places, we should make it a proper setting. 

It should probably update here also: https://github.com/readthedocs/ethical-ad-server/blob/005b2ddf3cd52c33c57bcba458b3f9cb9ce857ba/adserver/analyzer/tasks.py#L168

ericholscher (approved):

PR #1044, opened 2025-06-26
Add a base analyzer for Trafilatura

davidfischer:


ericholscher (approved):
Feels harder to review this without the refactor of the original.. Should likely have just been in the same PR as https://github.com/readthedocs/ethicalads-ext/pull/46/

davidfischer:
> Feels harder to review this without the refactor of the original.. Should likely have just been in the same PR as https://github.com/readthedocs/ethicalads-ext/pull/46

Do you want me to just close this and take the code private? If we want to use `trafilatura`, seems useful to have a base class for it though.

PR #1048, opened 2025-07-01
Add a connection max age for the replica

davidfischer:
This was an oversight I noticed while investigating some DB slowness.


ericholscher (approved):
Oh yea -- if we're re-opening a connection each time, that could be a huge speed hit.

davidfischer:
After we upgrade to Django 5.2, we should also try the [built-in pooling](https://docs.djangoproject.com/en/5.2/ref/databases/#connection-pool)

PR #1052, opened 2025-07-15
Fix a number of warnings

davidfischer:
- Warnings from allauth (eg. `settings.ACCOUNT_EMAIL_REQUIRED is deprecated, use: settings.ACCOUNT_SIGNUP_FIELDS = ['email*', 'password1*', 'password2*']`)
- Changes with Django's storages API which will become errors in Django 5.1
- Rely on the builtin `fromisoformat`

### DIFF
--- a/adserver/management/commands/archive_offers.py
+++ b/adserver/management/commands/archive_offers.py
@@ -15,7 +15,7 @@ import tempfile
 from pathlib import Path
 
 from django.conf import settings
-from django.core.files.storage import get_storage_class
+from django.core.files.storage import storages
 from django.core.management import CommandError
 from django.core.management.base import BaseCommand
 from django.db import connections
@@ -135,7 +135,7 @@ class Command(BaseCommand):
 
     def copy_offer_dump(self, archive_filepath):
         """Copy offer CSV files to settings.BACKUPS_STORAGE."""
-        if not hasattr(settings, "BACKUPS_STORAGE"):
+        if "backups" not in settings.STORAGES:
             self.stdout.write(
                 self.style.WARNING(
                     _(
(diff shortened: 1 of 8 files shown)


adserver/management/commands/archive_offers.py
```diff
@@ -135,7 +135,7 @@ def handle_archive_day(self, day):
 
     def copy_offer_dump(self, archive_filepath):
         """Copy offer CSV files to settings.BACKUPS_STORAGE."""
-        if not hasattr(settings, "BACKUPS_STORAGE"):
+        if "backups" not in settings.STORAGES:
```
ericholscher, line 138: Not sure we're even using this file anymore? I guess we can keep it around for now.

config/settings/base.py
```diff
@@ -389,10 +389,9 @@
 # --------------------------------------------------------------------------
 ACCOUNT_ADAPTER = "adserver.auth.adapters.AdServerAccountAdapter"
 ACCOUNT_USER_MODEL_USERNAME_FIELD = None
-ACCOUNT_EMAIL_REQUIRED = True
-ACCOUNT_USERNAME_REQUIRED = False
 ACCOUNT_LOGIN_ON_PASSWORD_RESET = True
 ACCOUNT_MAX_EMAIL_ADDRESSES = 1
+ACCOUNT_SIGNUP_FIELDS = ["email*", "password1*", "password2*"]
```
ericholscher, line 394: The * here seems a bit magical -- worth a comment, or a link to the docs?

ericholscher (approved):

adserver/management/commands/archive_offers.py
davidfischer: We can probably delete the whole management command. We're not using it.

config/settings/base.py
davidfischer: Added

PR #1056, opened 2025-08-05
Ad images not going through CDN/custom domain

davidfischer:
We appear to have lost this feature in a past Django or storages upgrade.

Ref: https://github.com/readthedocs/ethical-ad-client/issues/228


ericholscher (approved):
Aye.. I feel like this stuff gets lost when we upgrade and it's super hard to catch.

PR #1057, opened 2025-08-08
Improve UX around renewing flights

davidfischer:
- Add a button to renew completed flights
- Change verbiage to "renew"

### DIFF
--- a/adserver/templates/adserver/advertiser/flight-list.html
+++ b/adserver/templates/adserver/advertiser/flight-list.html
@@ -31,7 +31,7 @@
   {% if has_advertiser_edit_permission %}
   <a href="{% url 'flight_request' advertiser.slug %}" class="btn btn-sm btn-outline-info mb-3" role="button" aria-pressed="true">
     <span class="fa fa-plus mr-1" aria-hidden="true"></span>
-    <span>{% trans 'Request a new flight' %}</span>
+    <span>{% trans 'Renew or request a new flight' %}</span>
   </a>
   {% endif %}
 </aside>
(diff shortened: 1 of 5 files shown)


adserver/templates/adserver/advertiser/flight-list.html
```diff
@@ -31,7 +31,7 @@ <h1>{% block heading %}{% trans "Manage advertising flights" %}{% endblock headi
   {% if has_advertiser_edit_permission %}
   <a href="{% url 'flight_request' advertiser.slug %}" class="btn btn-sm btn-outline-info mb-3" role="button" aria-pressed="true">
     <span class="fa fa-plus mr-1" aria-hidden="true"></span>
-    <span>{% trans 'Request a new flight' %}</span>
+    <span>{% trans 'Renew or request a new flight' %}</span>
```
ericholscher, line 34: This feels a little hacky to note both here -- I don't know if a single bit of feedback is worth updating all the UI items? 

Do we want something like `Set up a new flight`, since that's what is actually happening?

ericholscher (commented):
I think we want:

* The ability for the user to enable auto-renewing of the current flight, which will auto-create and start a new flight at the end of the current
* The ability for them to set up a new flight that we will approve, possibly changing ads or targeting

I wonder about removing the 10% discount for auto-renewal, since it seems like that is what is stopping us from allowing folks to opt in to auto-renew automatically, since it changes pricing and we don't dynamically price yet?

davidfischer:
> I guess the bigger thing is trying to get our pricing to automatically update based on flight params, but that's a larger project. I really do think we want the calculator functionality in the dashboard tho, and let users basically price and prepare a flight and just hit "submit for approval" at the end of that process knowing the cost.

The first bits of this are done. The dashboard stores our prices. The problem is a lot of real campaigns fall in some middle ground that the calculator doesn't handle. For example, they target a subset of countries in a region or they target multiple topics. It's tricky. One obvious thing to add is our standard discount thresholds.

PR #1067, opened 2025-09-11
Add a periodic task to update domain centroids

davidfischer:
In production, this should take sub 1 minute


ericholscher (approved):
Simple enough.

PR #1206, opened 2026-04-15
Pin prompt-toolkit>=3.0.43 to fix TypeError on Python 3.12

ericholscher:
## Summary
- Older versions of `prompt-toolkit` pass bare coroutines to `asyncio.wait()`, which raises `TypeError` on Python 3.12+ (coroutines were deprecated in 3.8, became an error in 3.12)
- `prompt-toolkit` is a transitive dependency via `celery` → `click-repl`
- Pins `prompt-toolkit>=3.0.43` which wraps coroutines with `ensure_future()` before passing to `asyncio.wait()`

## Error
```
File "prompt_toolkit/renderer.py", line 514, in wait_for_cpr_responses
    await wait(coroutines, return_when=FIRST_COMPLETED)
TypeError: Passing coroutines is forbidden, use tasks explicitly.
```

## Test plan
- [ ] Verify celery starts without the `TypeError` on Python 3.12

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,6 +12,10 @@ dependencies = [
     # Celery <5.5 has a stability issue the redis broker: https://github.com/celery/celery/issues/8030
     "celery[redis]==5.5.0rc1",
 
+    # Celery depends on prompt-toolkit via click-repl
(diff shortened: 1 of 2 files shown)


davidfischer (approved):
I built the image and verified celery started and ran correctly with this change.

PR #1209, opened 2026-04-18
Improve placement priority performance

davidfischer:
Looking at the performance degradation in NR, the problem didn't seem to be a "query" problem in so far as NR didn't show elevated database times. Looking at the code, this led me to believe the degradation (~20ms avg) from the previous change was due to python deserializing too many objects from the database.

After discussion with the ol' AI (Gemini), this change should avoid deserializing a pretty large number of objects and be similar to what we had before. Previously, we were requesting every ad and its ad types for each ad from each candidate flight. Assuming 3 ad types per ad and ~30 flights and ~4 ads per flight, we're looking at ~350 more objects being deserialized from the DB. The query itself didn't appear more expensive, but turning that data into Django model objects was. That's at least the working theory.

I have verified that this change looks good from a query perspective and I reviewed every query in the critical path.

## References

- [Django Case/When](https://docs.djangoproject.com/en/5.2/ref/models/conditional-expressions/)
- These case/when's result in Postgres [CASE statements](https://www.postgresql.org/docs/current/functions-conditional.html#FUNCTIONS-CASE)

### DIFF
--- a/adserver/decisionengine/backends.py
+++ b/adserver/decisionengine/backends.py
@@ -14,7 +14,6 @@ from ..constants import COMMUNITY_CAMPAIGN
 from ..constants import HOUSE_CAMPAIGN
 from ..constants import PAID_CAMPAIGN
 from ..constants import PUBLISHER_HOUSE_CAMPAIGN
-from ..models import Advertisement
 from ..models import Flight
 from ..models import Region
 from ..models import Topic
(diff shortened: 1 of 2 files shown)


davidfischer:
I'd still like a review here but I'm going to test a version of this out.

davidfischer:
After deploying this to test it out, performance looked normal. Glad to finally have this!

ericholscher:
I looked over this, and it looks good. Even better knowing that it works in prod :D

PR #1211, opened 2026-04-21
Test multi-ad-type placement priority weighting logic

ericholscher:
## Summary
Added a comprehensive test case to verify that flights with ads matching multiple placements are weighted by the maximum priority across all matching placements, rather than just the first match.

## Key Changes
- Added `test_multi_ad_type_placement_priority()` test method that:
  - Creates two placements with different priorities (1 and 10)
  - Sets up a flight with an ad matching both placement types (ad_types a and b)
  - Sets up a comparison flight with an ad matching only the low-priority placement (ad_type a)
  - Verifies through probabilistic selection that the multi-type ad flight is selected ~10x more often, confirming it uses the maximum priority (10) rather than the first match priority (1)
  - Uses 1000 iterations to ensure statistical significance of the weighting behavior

## Implementation Details
- The test validates that when an advertisement's ad_types match multiple placements, the flight's weight calculation uses the maximum priority value across all matching placements
- Base weights are equal between the two flights, so any selection ratio difference directly reflects the placement priority weighting logic
- The assertion `self.assertGreater(ratio, 5.0)` confirms the ratio is significantly higher than 1.0, validating the max-priority semantics over first-match semantics

https://claude.ai/code/session_015nHNTeYYgHcPYpR7QSJWzk

### DIFF
--- a/adserver/tests/test_decision_engine.py
+++ b/adserver/tests/test_decision_engine.py
@@ -699,6 +699,86 @@ class DecisionEngineTests(TestCase):
         # Used to be ~1, now should be ~10
         self.assertGreater(ratio, 5.0)
 
+    def test_multi_ad_type_placement_priority(self):
+        # An ad whose ad_types match multiple placements should weight its flight
+        # by the MAX priority across all matching placements, not the first match.
+        for flight in Flight.objects.all():
+            flight.live = False
+            flight.save()
+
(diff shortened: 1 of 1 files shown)


davidfischer (approved):
This looks OK and is a more comprehensive test.

PR #1215, opened 2026-05-04
Applebot detection fix

davidfischer:
Applebot is not being properly detected as a bot.


ericholscher (approved):

ericholscher:
Pretty sure CF probably has better bot detection than we do? I wonder if we can lean on it.

davidfischer:
> Pretty sure CF probably has better bot detection than we do? I wonder if we can lean on it.

We do send the "bot score" down to the application in a header. Perhaps worth checking.

PR #1216, opened 2026-05-04
Notify on large daily publisher traffic spikes

davidfischer:
- 200% increases on a publisher over a single day (posts to slack)
- Keep the weekly notification

### DIFF
--- a/adserver/tasks.py
+++ b/adserver/tasks.py
@@ -1270,6 +1270,107 @@ def notify_of_completed_flights():
                 message.send()
 
 
+def check_publisher_metrics_change(
+    publisher,
+    metric,
+    current_start,
+    current_end,
+    previous_start,
+    previous_end,
+    difference_threshold,
+    min_views,
+    current_period_name,
+    previous_period_name,
+    check_increase_only=False,
+    min_views_in_current_only=False,
+):
+    """
+    Check if a publisher's metrics have changed significantly between two periods
+    and send a Slack notification if they have.
+
+    :param publisher: The Publisher instance to check
+    :param metric: The metric to compare (e.g. "views", "revenue")
+    :param current_start: Start date for the current period
+    :param current_end: End date for the current period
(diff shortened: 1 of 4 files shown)


adserver/tasks.py
```diff
@@ -1270,6 +1270,107 @@ def notify_of_completed_flights():
                 message.send()
 
 
+def check_publisher_metrics_change(
```
ericholscher, line 1273: This feels pretty janky. Presumably some of these should just live in the caller and it can send the slack message with the various variables?

ericholscher:
I also think the weekly notifications are probably too low of a threshold -- anything that alerts regularly with ~15 issues is going to get ignored.

adserver/tasks.py
davidfischer: Are you proposing having 2 tasks without reusing code? I'm not sure what you're proposing here.

davidfischer:
> I also think the weekly notifications are probably too low of a threshold -- anything that alerts regularly with ~15 issues is going to get ignored.

I tend to agree. Currently weekly notifications require a minimum of 10k with a difference of 25%. What do you think? Go to 33%? 50%?

adserver/tasks.py
ericholscher: Just using fewer variables, and keeping more of the task specific ones in the tasks, unless we expect to run a large number of these.

davidfischer:
Ok, I simplified the utility function a bit. I made the human readable parts not part of the function at all and the actual slack messaging in the tasks themselves.

I also changed the default weekly threshold 25% -> 33%.

ericholscher (approved):

ericholscher:
> I also changed the default weekly threshold 25% -> 33%.

I'd probably scale it based on size? Like, I want to know about 33% for Daily.dev, but not jsbin. We don't need to block this PR on that, though.

PR #1222, opened 2026-05-15
Exclude $ from copied amount in staff payout finish view

ericholscher:
Move the dollar sign outside the span that the copy button reads from
so only the numeric amount is copied to the clipboard.


davidfischer (approved):
Change looks good and I was able to fix up the test.

PR #1236, opened 2026-06-22
Link the dashboard back to the API docs and website (settings-driven)

ericholscher:
Human summary: I found a few places where linking to the website / API docs would be useful. Had AI through together a quick PR. 

Follow up: Need to add the ENV on deploy to get the website link. 

## What

Adds links from the dashboard back to the API reference and marketing website "where reasonable," and makes those URLs settings-driven so self-hosted instances can point them at their own resources (or leave them unset to hide the links).

## Why

The dashboard only linked back to external resources in a few ad-hoc, hardcoded places. The clearest gap was the API token page, which explains tokens are used to access the API but never linked to the API docs.

The readthedocs site is API/developer documentation rather than end-user help, so the links are scoped to the "API folks" audience — there's no general "Documentation" link on the user-facing dashboard.

## Changes

**New settings** (in `config/settings/base.py`, exposed via the existing settings context processor):

| Setting | Default |
|---|---|
| `ADSERVER_API_DOCUMENTATION_URL` | `…/en/latest/user-guide/api.html` |
| `ADSERVER_WEBSITE_URL` | unset (link hidden) |

**Where the links appear:**
- **Footer** — API and Website links, each rendered only when its URL is configured (mirrors the existing privacy/publisher-policy footer links). `ADSERVER_WEBSITE_URL` is unset by default, so the website link only appears once the env var is set on deploy.
- **API token page** — links to the API documentation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/adserver/templates/adserver/accounts/api-token.html
+++ b/adserver/templates/adserver/accounts/api-token.html
@@ -22,6 +22,10 @@
 
 <p>{% trans 'API tokens allow you to use the ad server API to download reports.' %}</p>
 
+{% if adserver_api_documentation_url %}
+  <p>{% blocktrans %}See the <a href="{{ adserver_api_documentation_url }}" target="_blank">API documentation</a> for the available endpoints and usage details.{% endblocktrans %}</p>
+{% endif %}
+
 {% if not object_list %}
   <form method="post" action="{% url 'api_token_create' %}">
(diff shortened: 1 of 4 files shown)


davidfischer (approved):
Looks good 🚢

PR #1246, opened 2026-07-23
Add ability to ratelimit ASNs

davidfischer:
This allows us to add ratelimits to selected ASNs with a setting/envvar. Cloudflare is already sending ASN down to the application.


ericholscher (approved):
Seems sadly useful.

PR #1251, opened 2026-08-03
Add CORS to the decision endpoint

davidfischer:
Ref: https://github.com/readthedocs/ethical-ad-client/issues/242

### DIFF
--- a/adserver/tests/test_api.py
+++ b/adserver/tests/test_api.py
@@ -292,6 +292,36 @@ class AdDecisionApiTests(BaseApiTest):
         resp_json = resp.json()
         self.assertEqual(resp_json["id"], "ad-slug", resp_json)
 
+    def test_cors_headers(self):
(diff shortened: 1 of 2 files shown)


ericholscher:
Do we actually want CORS on this? That issue seemed like it was opened by an AI bot, and I'm not really sure what the real use case is.

davidfischer:
> Do we actually want CORS on this? That issue seemed like it was opened by an AI bot, and I'm not really sure what the real use case is.

Overall, I think this is pretty unimportant. I realize the issue was opened by an AI bot. Technically, the effective change here is that people could use the JSON endpoint directly instead of using JSONP. Overall, I think that's reasonable but pretty low priority.

ericholscher (approved):
I feel like this might make scams a bit easier to pull off, but I guess in the end it's not a huge difference between json/jsonp and could make some things cleaner.

